PR #4, opened 2022-04-05
Allocation and Allocation Groups

gnomadic:
This PR introduces the Chef, which will handle distribution of the initial supply of METRIC.  The Chef accomplishes this by defining "allocation groups", which each have a set amount of shares.  Over time, METRIC is emiitted to these groups at a rate defined by the amount of shares they hold.

This code is a "soft fork" of Sushiswap's MasterChef, leveraging their design and math with a new, simpler data structure.


ftfdfrancoeur65 (commented):
few comments on how we are storing/editing AllocationGroup

contracts/Chef.sol
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ In sushi's master chef, the design allows the controller to submit LP contracts for yield farming, and then user's can stake their LP tokens.
+
+ In this contract, there is no concept of a user staking their LP tokens - and instead of LP contract, the controller is submitting Allocation Groups.
+
+ So in sushi:
+
+ 1.  Every `x` blocks, calculate minted Sushi Tokens for each LP contract based on their (shares / total shares)
+ 2.  Then, do the math to figure out how many rewards each LP token is worth (based on the total amount of LP tokens staked)
```
yahehe, line 25: Why is the Sushi algorithm being described here? Feels like good info for the PR description but not needed to document _this_ code?

yahehe:
PR description?

contracts/Chef.sol
```diff
@@ -0,0 +1,198 @@
...
+        // TODO confirm budget is correct with assertions
+        // Not sure we can project emission rate over X years?
+        // Not entirely sure how to handle this, but we can at least try to make it work.
+        // ^^ will help with fuzz testing
+
+        uint256 blocks = block.number.sub(_lastRewardBlock);
+
+        uint256 accumulated = blocks.mul(METRIC_PER_BLOCK);
+
+        _accMETRICPerShare = _accMETRICPerShare.add(accumulated.mul(ACC_METRIC_PRECISION).div(_totalAllocPoint));
```
yahehe, line 137: Would we save any gas by pre-calculating precision/totalAlloc?

contracts/Chef.sol
```diff
@@ -0,0 +1,198 @@
...
+        _lastRewardBlock = block.number;
+    }
+
+    function harvestAll() public {
+        for (uint256 i = 0; i < _allocations.length; i++) {
```
yahehe, line 142: Is there an upper bound to _allocations.length?

contracts/Chef.sol
```diff
@@ -0,0 +1,198 @@
...
+            harvest(i);
+        }
+    }
+
+    function harvest(uint256 agIndex) public {
+        AllocationGroup storage group = _allocations[agIndex];
+
+        updateAccumulatedAllocations();
```
yahehe, line 150: Should this take `group` as a param?

contracts/Chef.sol
```diff
@@ -0,0 +1,198 @@
...
+
+        uint256 pending = group.shares.mul(_accMETRICPerShare).div(ACC_METRIC_PRECISION).sub(group.rewardDebt);
```
yahehe, line 152: same question about pre-calculating constants to conserve gas

diagrams/README.md
```diff
@@ -0,0 +1,62 @@
+# Design
+
+
+## Core Assumptions
+
+
+1.  Supply is fixed at 1B
+2.  Rewards will never end
```
yahehe, line 8: If supply is fixed, can we guarantee rewards never end?

diagrams/README.md
```diff
@@ -0,0 +1,62 @@
...
+4.  Reuse the same code as the vesting contract (makes the audit cheaper/easier)
+5.  Payment splitter – uses shares: https://docs.openzeppelin.com/contracts/2.x/api/payment 
+6.  Define addresses and their number of shares, have a payment function that distributes
+
+
+### Faucet
+
+1.  Want faucet to have a balance
+2.  Needs to be replenished
+3.  Cron job triggers distribution from rewards contract daily?
```
yahehe, line 42: Is this cron centralized?

contracts/Chef.sol
```diff
@@ -0,0 +1,227 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "hardhat/console.sol";
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "./MetricToken.sol";
+import "hardhat/console.sol";
```
gnomadic, line 8: don't check this in!

diagrams/README.md
gnomadic: right now `METRIC_PER_BLOCK` tracks the emission rate, and it's hardcoded to 4 - but when we get into the Rewards design we will want to put this behind a function which will dynamically calculate the emission rate - OR - we will want to invalidate these assumptions.  It's much more common to have a projected supply over the years and emit based on the network's block-rate to match that projection as close as possible.

Either way, once we get into rewards, we can revisit these!

diagrams/README.md
gnomadic: if we go down this route, yes, it would be - flipside would host the cron and foot the gas fees for it.  

Another option is to add to the gas cost of end-users so that they will be calling the update and distribute functions whenever they do anything - or some sort of round-robin where a random (unlucky?) person gets the bill.

contracts/Allocator.sol
```diff
@@ -0,0 +1,186 @@
...
+contract Allocator is AccessControl {
+    using SafeMath for uint256;
+
+    bytes32 public constant ALLOCATION_ROLE = keccak256("ALLOCATION_ROLE");
+
+    uint256 public METRIC_PER_BLOCK = 4 * 10**18;
+    uint256 public constant ACC_METRIC_PRECISION = 1e12;
+
+    bool private _rewardsActive;
+    AllocationGroup[] private _allocations;
```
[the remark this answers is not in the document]
gnomadic: can we talk that through?  We need to iterate over all the allocation groups (either in the contract itself or in the Dapp) - and iterating over maps isn't super convenient

contracts/Chef.sol
gnomadic: this is pretty common with smart contracts because the code and its comments will be available on etherscan (or the chain's fork).  I also think it's pretty web3-style because it means the comments are decentralized and immutable as well as the code.

I will pull it into the readme, and I expect us to have a gitbook at some point where it can also live.  Once our documentation source is finalized (assuming before launch) we can link to that from the smart contract

contracts/Chef.sol
gnomadic: no but we expect 4 so an implicit upper bounds of 10 or so - I dropped that to a `uint8` -> but if/when we use this as a foundation for the staking implementation, we will want to use uint256 because the "stakers" will be arbitrary users.

contracts/Chef.sol
gnomadic: interesting - so yes, but it's negligible (100 runs with an average of less than 100 gas saved) - however adding a new global variable and modifying the 3 AG methods to calculate it brings the deployment cost from` 2280664` to `2331879`

contracts/Chef.sol
gnomadic: I mean we would still want to do a lookup for the share value, debt, etc within the contract because we can't trust the client to pass in their group and it's values accurately.

ftfdfrancoeur65 (approved):
looks good to me

contracts/Chef.sol
yahehe: The code being commented is fine, I'm confused why the comments are describing _other_ code. Fine if you're pulling that out, I'd just expect "We started with this other thing someone else wrote, so if you're interested you can look there, but we here we describe our thing"

diagrams/README.md
yahehe: Will rewards be figured out before these contracts are deployed?

diagrams/README.md
yahehe: Would `update` and `distribute` be expensive for the end user if they were only calculated for the end user?

Fine for the cron to be centralized as long as others can choose to call it if they so desire (e.g. if Flipside goes under or stops caring about the cron)

contracts/Chef.sol
gnomadic: yes -- see that it's tagged `public`, this is intended to be called from an external source (maybe another contract, maybe a user through a UI)

diagrams/README.md
gnomadic: yes - we need to get 80% of the implementation ready to even be able to book an auditor - and then they'll take at least a month before they can even get started --

Launch will include the token, vesting, staking, the faucet, and question creation

diagrams/README.md
gnomadic: `update` is `updateAccumulatedAllocations()` in this contract, and `distribute` is `harvest()` btw -- the update function applies to everyone although harvest will only harvest for the current user.   Expensive is relative, but they do have gas fees associated with em.

And yes - Update is called pretty regularly through this contract, and is a public function anyone can call.

contracts/Chef.sol
gnomadic: hmmm.  Good point!  the original intention was that I had to wrap my head around sushi's masterchef completely to be able to modify it so I could write this one - so I was partially writing those comments to help my understanding.

But now, we have our own implementation so this does make sense.

PR #7, opened 2022-04-20
adjustments to Allocator

ftfdfrancoeur65:


gnomadic:
We've been doing a deep-dive review with @jfmyers and @addyxvii  and have made a ton of changes to the allocator since this went in (I've even renamed it to `Chef` (even though we're not minting SUSHI as a chef would, it's still based on a Chef contract) 

These changes are awesome.  I'll port them over to the new implementation once the OG PR is merged in, because manually merging this one in is going to be tedious

PR #19, opened 2022-05-06
create a base deployment script for the protocol

ftfdfrancoeur65:
- create deployment script for the protocol (MetricToken, Chef)
- introduce new hardhat plugins for deployment/verification


hardhat.config.ts
```diff
@@ -35,12 +36,20 @@ const config: HardhatUserConfig = {
     currency: "USD",
   },
   etherscan: {
-    apiKey: process.env.ETHERSCAN_API_KEY,
+    apiKey: {
+      ropsten: process.env.ETHERSCAN_API_KEY || "",
```
gnomadic, line 40: should these be `ETHERSCAN_ROPSTEN_API_KEY` etc?  Also maybe add em to .env.template if we got one?

deploy/00_deploy_protocol.js
```diff
@@ -0,0 +1,40 @@
...
+
+  try {
+    await hre.run("verify:verify", {
+      address: chef.address,
+      constructorArguments: [metricToken.address],
+      contract: "src/contracts/Chef.sol:Chef",
+    });
+  } catch (error) {
+    console.log("error:", error.message);
+  }
```
gnomadic, line 38: a convention @jdaly13 introduced me to was to write the addresses to a json file after everything has successfully been deployed - what do ya think?  we could have a {network-name}-address.json file that we keep in source control so we always have em

deploy/00_deploy_protocol.js
```diff
@@ -0,0 +1,40 @@
+// deploy/00_deploy_my_contract.js
```
jdaly13, line 1: I apologize as I'm just catching up on all this - I'm assuming this file is just middleware to deploy two different contracts but don't see the initial file - when you deploy does it also await for .deployed() as well?
In terms of writing contract to json would be something like this
```
const address = {};
address.metricToken = metricToken.address
address.chef = chef.address

  fs.writeFile(
    `${process.env.network}-address.json`,
    JSON.stringify(address),
     (err) => {
      if (err) throw err;
      console.log("address file created");
    }
  );
  ```

Comment (ftfdfrancoeur65) [reply deploy/00_deploy_protocol.js:1]: hmm, so this file can be run using a hardhat plugin that introduces a new hardhat command `deploy`

Try running this `npx hardhat deploy --network ropsten` and you'll see a folder (which i added to gitignore) with all the deployment info! `deployments/ropsten`

Comment (ftfdfrancoeur65) [reply deploy/00_deploy_protocol.js:1]: https://github.com/wighawag/hardhat-deploy :)

File: .gitignore [hunk 4]
```diff
@@ -3,6 +3,7 @@ node_modules
 coverage
 coverage.json
 typechain
+deployments
```
ftfdfrancoeur65, line 6: @jdaly13 @gnomadic this is the folder where you'll see all the contract addys.

My thought is that we could maybe get a bit more specific on this and just ignore test files, while adding mainnet deployments information.

deploy/00_deploy_protocol.js
jdaly13: ah that makes sense - will look into wasn't aware of it

deploy/00_deploy_protocol.js
```diff
@@ -0,0 +1,40 @@
...
+  const { deploy } = deployments;
+  const { deployer } = await getNamedAccounts();
+  console.log("deployer:", deployer);
+  const metricToken = await deploy("MetricToken", {
+    from: deployer,
+    args: [deployer],
+    log: true,
+  });
+
+  const chef = await deploy("Chef", {
```
jdaly13, line 14: will have to change to TopChef as Chef will be an abstract contract that others will inherit from

.gitignore
jdaly13: addys? Yeah I would say ignore subfolders but keep deployments?

jdaly13 (commented):
This looks good to me pending the changes from Chef to TopChef - thank you for your work on this

PR #26, opened 2022-05-19
25 create abstract chef contract in which others can inherit from

jdaly13:
goal of this PR is to create a base class called Chef in which other chefs will inherit from - also some slight refactoring

### DIFF
--- a/src/contracts/Chef.sol
+++ b/src/contracts/Chef.sol
@@ -1,25 +1,48 @@
 //SPDX-License-Identifier: Unlicense
 pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
 import "./MetricToken.sol";
 
-abstract contract Chef {
-    using SafeMath for uint256;
-    uint256 public _metricPerBlock;
+abstract contract Chef is Ownable {
+    uint256 private _metricPerBlock;
     uint256 public constant ACC_METRIC_PRECISION = 1e12;
 
-    bool private _rewardsActive; // DO WE MAKE ubiquitous per each chef? 
+    bool private _rewardsActive; // DO WE MAKE ubiquitous per each chef?
+    uint256 private _lastRewardBlock; 
 
     MetricToken private _metric;
 
-    function setMetricPerBlock(uint256 metricAmount) public {
+    //------------------------------------------------------Setters
+
+    function toggleRewards(bool isOn) public onlyOwner() {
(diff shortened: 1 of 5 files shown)


src/contracts/Chef.sol
```diff
@@ -1,203 +1,45 @@
...
+    function toggleRewards(bool isOn) public onlyOwner() {
         _rewardsActive = isOn;
         _lastRewardBlock = block.number;
     }
 
-    //------------------------------------------------------Getters
-
-    function getAllocationGroups() public view returns (AllocationGroup[] memory) {
-        return _allocations;
+    function setMetricPerBlock(uint256 metricAmount) public virtual {
```
jdaly13, line 23: should this have the onlyOwner modifier?

addyxvii: I think so, don't want just anyone to be able to call this

src/contracts/Chef.sol
```diff
@@ -1,203 +1,45 @@
 //SPDX-License-Identifier: Unlicense
 pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/access/AccessControl.sol";
-import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
 import "./MetricToken.sol";
 
-// Heavily Inspired by Sushi's MasterChefv2 - but with a few changes:
```
gnomadic, line 8: can you move this to a readme?  the wall of text was how I worked out how it works and might be useful for future contributors (agree it shouldn't be in the file)

src/contracts/Chef.sol
```diff
@@ -209,16 +51,18 @@ contract Chef is AccessControl {
...
-    struct AllocationGroup {
-        address groupAddress;
-        uint256 shares;
-        bool autodistribute;
-        uint256 rewardDebt; // keeps track of how much the user is owed or has been credited already
-        uint256 claimable;
-    }
 }
+
+// DO WE ADD THESE TO MAIN CHEF OR PER TYPE OF CHEF
```
gnomadic, line 59: can ya start the first line `TODO` so it's easy to find?  also recommend the Todo extension for vscode https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree

gnomadic:
This is a good iteration.  I do think more functionality can live in the abstract base chef - for example I'd love to see `updatedAccumulatedAllocations` in there -- BUT let's an iteration of the Staking contract in place, and we can find the common pieces of code/logic and abstract them as possible

gnomadic (approved):

PR #28, opened 2022-05-24
remove metric from topchef add getter setter in chef base

jdaly13:
realized I left out metric token out of base chef - these changes call a setter within constructor only by owner and a getter as well to send transfers

### DIFF
--- a/src/contracts/Chef.sol
+++ b/src/contracts/Chef.sol
@@ -36,7 +36,7 @@ abstract contract Chef is Ownable {
         _metricPerBlock = metricAmount * 10**18;
     }
 
-    function setLastRewardBlock(uint256 blockNumber) public virtual {
+    function setLastRewardBlock(uint256 blockNumber) internal virtual {
         _lastRewardBlock = blockNumber;
     }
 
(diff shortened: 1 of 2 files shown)


src/contracts/TopChef.sol
```diff
@@ -146,7 +144,8 @@ contract TopChef is Chef {
         require(group.claimable != 0, "No claimable rewards to withdraw");
         // TODO do we want a backup in case a group looses access to their wallet
         require(group.groupAddress == _msgSender(), "Sender does not represent group");
-        _metric.transfer(group.groupAddress, group.claimable);
+        MetricToken metric = getMetricToken();
```
jdaly13, line 147: Should we be using SafeERC20 package to use SafeERC20.safeTransfer

jdaly13:
do we want ability to destroy contract? @gnomadic @addyxvii

gnomadic:
I don't think so right? I can't imagine a use case

src/contracts/Chef.sol
```diff
@@ -54,6 +58,10 @@ abstract contract Chef is Ownable {
         return _rewardsActive;
     }
 
+    function getMetricToken() internal view virtual returns (MetricToken) {
```
jdaly13, line 61: not sure if making this internal is necessary but thought only the contracts derived from this should be calling it

src/contracts/Chef.sol
```diff
@@ -66,4 +74,8 @@ abstract contract Chef is Ownable {
     event Harvest(address harvester, uint256 agIndex, uint256 amount);
     event Withdraw(address withdrawer, uint256 agIndex, uint256 amount);
 
+    receive() external payable virtual {}
+
+    function withdrawMoney() public onlyOwner() {}
+
```
jdaly13, line 80: these two have blank implementations since an abstract contract cannot receive funds but those who inherit from can

addyxvii (approved):

PR #29, opened 2022-05-24
Create staking contract

addyxvii:
This PR is to add staking to the MDAO protocol


src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,158 @@
...
+
+//TODO StakeMetric, updateStaker, removeStaker, updateAccumulatedStakingRewards, harvest and Claim can be moved to Chef
+
+contract StakingChef is Chef {
+    using SafeMath for uint256;
+    Staker[] private _stakes;
+    uint256 private _totalAllocPoint;
+    uint256 private _lifetimeShareValue;
+
+    MetricToken private _metric;
```
jdaly13, line 16: so my PR has added Metric to base Chef

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,158 @@
...
+
+    constructor(address metricTokenAddress) {
+        _metric = MetricToken(metricTokenAddress);
```
jdaly13, line 19: hopefully should be able to utilize - `setMetricToken(metricTokenAddress);`

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,158 @@
...
+        setMetricPerBlock(4);
+    }
+
+    mapping(address => bool) public override addressExistence;
```
jdaly13, line 23: just curious can you override a mapping?

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,158 @@
...
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "./Chef.sol";
+import "./MetricToken.sol";
+
+//TODO StakeMetric, updateStaker, removeStaker, updateAccumulatedStakingRewards, harvest and Claim can be moved to Chef
+
+contract StakingChef is Chef {
+    using SafeMath for uint256;
+    Staker[] private _stakes;
+    uint256 private _totalAllocPoint;
```
jdaly13, line 13: I wonder if these two can be in Chef or we create functions in Chef for getters and setters which can be overiden

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,158 @@
...
+            rewardDebt: metricAmount.mul(_lifetimeShareValue).div(ACC_METRIC_PRECISION),
+            claimable: 0
+            });
+
+            _stakes.push(stake);
+            _totalAllocPoint = _totalAllocPoint.add(stake.metricAmount);
+            _metric.transfer(address(this), stake.metricAmount);
+    }
+
+    function updateAccumulatedStakingRewards() public {
```
jdaly13, line 96: Let's work on adding this functionality to Chef - honestly I would like to get a deeper understanding of this functionality as it lives here and TopChef at the moment

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+// --------------------------------------------------------------------- Manage rewards and Principal
+
+    function claim(uint256 stakeIndex) public {
+        Staker storage stake = _stakes[stakeIndex];
+
+        require(stake.claimable != 0, "No claimable rewards to withdraw");
+        require(stake.stakeAddress == _msgSender(), "Sender can not claim");
+        _metric.transfer(stake.stakeAddress, stake.claimable);
```
jdaly13, line 117: hoping to do something like 
```
MetricToken metric = getMetricToken();
SafeERC20.safeTransfer(IERC20(metric), stake.stakeAddress, stake.claimable);
                ```

File: src/contracts/StakingChef.sol [hunk 7]
```diff
@@ -0,0 +1,158 @@
...
+            stakeAddress: newAddress,
+            metricAmount: metricAmount, 
+            startDate: newStartDate,
+            rewardDebt: metricAmount.mul(_lifetimeShareValue).div(ACC_METRIC_PRECISION),
+            claimable: 0
+        });
+
+        _stakes.push(stake);
+        _totalAllocPoint = _totalAllocPoint.add(stake.metricAmount);
+        _metric.transfer(address(this), stake.metricAmount);
```
jdaly13, line 45: just realized _metric won't be transferred from the contract itself
`SafeERC20.safeTransferFrom(IERC20(metric), msg.sender, address(this), stake.metricAmount)`

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,184 @@
...
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "./Chef.sol";
+
+contract StakingChef is Chef {
+    using SafeMath for uint256;
+    Staker[] private _stakes;
+    uint256 private _totalAllocPoint;
+    uint256 private _lifetimeShareValue;
+    uint256 public UserStake;
+    MetricToken public metric;
```
gnomadic, line 14: instead of defining MetricToken in both the StakingChef and the <base>Chef, can we make the variable _metric in the <base>Chef accessible to subclasses?

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+// --------------------------------------------------------------------- Manage rewards and Principal
+
+    function claim(uint256 stakeIndex) public {
+        Staker storage stake = _stakes[stakeIndex];
+        harvest(stakeIndex);
+
+        require(stake.claimable != 0, "No claimable rewards to withdraw");
+        require(stake.stakeAddress == _msgSender(), "Sender can not claim");
```
gnomadic, line 114: require message is inverted -> "Only staker can claim?" 

But also when we switch from an array to a mapping this won't be a problem, so probably no biggie

src/contracts/StakingChef.sol
```diff
@@ -0,0 +1,184 @@
...
+
+        SafeERC20.safeTransfer(IERC20(metric), msg.sender, stake.claimable);
+        stake.claimable = 0;
+
+        emit Withdraw(msg.sender, stakeIndex, stake.claimable);
+    }
+
+    function withdrawPrincipal(uint256 stakeIndex) public {
```
gnomadic, line 122: should the call to removeStaker call this?  BAH another thing that will be less confusing once we get the mapping in place.  Let's do that next

src/contracts/StakingChef.sol
jdaly13: @gnomadic @addyxvii - yes that makes sense - I have updated my PR to reflect this

jdaly13 (approved):

PR #31, opened 2022-05-26
move more methods to Chef

jdaly13:


test/chef-test.js
```diff
@@ -237,6 +237,18 @@ describe("Allocator Contract", function () {
 
       expect(BN(claimable)).to.equal(0);
     });
+
+    it("update allocationssssss", async function () {
```
addyxvii, line 241: Was this written by a 🐍?

jdaly13: Whooooooopsie!!!! I'll update

src/contracts/Chef.sol
```diff
@@ -55,6 +77,27 @@ abstract contract Chef is Ownable {
...
+
+    function getTotalAllocationShares() public view returns (uint256) {
+        return _totalAllocShares;
+    }
+
+    function getLifetimeShareValue() public view returns (uint256) {
+        return _lifetimeShareValue;
+    }
+
+    function accumulatedMetricDividedByShares(uint256 accumulatedWithPrecision) public view returns (uint256) {
```
gnomadic, line 97: is it possible to abstract away the ACC_METRIC_PRECISION from subclasses?  For example, could this function just take in accumulated metric and do the call to `getAcculatedWithmetricPrecision`?  It might be nice if the stakingChef/etc didn't have to know or care about this precision stuff

src/contracts/Chef.sol
```diff
@@ -41,6 +45,24 @@ abstract contract Chef is Ownable {
         return MetricToken(metricTokenAddress);
     }
 
+    function setLifetimeShareValue() public virtual {
+        uint256 accumulated = getAccumulated();
+        uint256 accumulatedWithMetricPrecision = getAcculatedWithmetricPrecision(accumulated);
+        _lifetimeShareValue = _lifetimeShareValue.add(accumulatedMetricDividedByShares(accumulatedWithMetricPrecision));
+    }
+
+    function setTotalAllocShares(uint256 shares) internal virtual {
```
gnomadic, line 54: can we rename this to `addTotalAllocShares`?  reading the subclass I had to come back and check the implementation 🙃

jdaly13: done

src/contracts/Chef.sol
jdaly13: I tried to make it modular setLifetimeShareValue - just calls 3 different functions within it so it's just a shortcut I guess
but you could call 
getAccumulated then whatever is returned from that call getAcculatedWithmetricPrecision and whatever is returned from that call accumulatedMetricDividedByShares and you can then take that value and add it to existing 
_lifetimeShareValue

gnomadic (approved):

addyxvii (approved):

PR #34, opened 2022-06-01
24-Question scaffold

gnomadic:
This PR introduces the Question Factory, Bounty Question, State Controller, and Claim Controller.

The Question Factory acts as the "public API if you will" for Question Management. 

The Bounty Question is a VERY simple NFT - arguably this is not upgradeable because it tracks a given question.

The ID of a bounty question is used by the state and claim controllers to track, well, the state of the question and the claims of the question.


src/contracts/Question/QuestionFactory.sol
```diff
@@ -0,0 +1,45 @@
...
+    function createQuestion(string memory uri, uint256 claimLimit) public payable {
+        uint256 newTokenId = _question.safeMint(_msgSender(), uri);
+        _questionStateController.initializeQuestion(newTokenId);
+        _claimController.initializeQuestion(newTokenId, claimLimit);
+    }
+
+    function upvoteQuestion() public payable {}
+
+    function claimQuestion() public payable {}
+
```
addyxvii, line 33: Why are these functions payable? won't they only be accepting METRIC?

addyxvii (approved):

PR #35, opened 2022-06-03
Switch to mapping

addyxvii:
This PR removes indexes from the staking contract to switch to mapping


gnomadic:
do you think it'll be possible to merge the allocationGroup and Staker structs... and then pull the mapping into the abstract chef?

src/contracts/StakingChef.sol
```diff
@@ -7,82 +7,64 @@ import "./Chef.sol";
...
-// --------------------------------------------------------------------- staking functions
-    function stakeMetric(
-        address newAddress,
-        uint256 metricAmount,
-        uint256 newStartDate
-    ) external nonDuplicated(newAddress) {
-    if (areRewardsActive() && getTotalAllocationShares() > 0) {
+
+    // --------------------------------------------------------------------- staking functions
+    function stakeMetric(uint256 metricAmount, uint256 newStartDate) external nonDuplicated(msg.sender) {
```
jdaly13, line 20: Think we should be using block.timestamp or at least do a require statement that newStartDate should be greater than block.timestamp

src/contracts/StakingChef.sol
```diff
@@ -7,82 +7,64 @@ import "./Chef.sol";
...
+    function removeStaker() external {
         if (areRewardsActive() && getTotalAllocationShares() > 0) {
             updateAccumulatedStakingRewards();
         }
-        removeAllocShares(_stakes[stakeIndex].metricAmount);
-
-        _stakes[stakeIndex] = _stakes[_stakes.length - 1];
-        _stakes.pop();
+        removeAllocShares(staker[msg.sender].metricAmount);
+        //Do we need to do anything else here to prevent removed staker from getting rewards?
```
jdaly13, line 49: Hmmm don't we need to refund metric amount to sender?

src/contracts/StakingChef.sol
```diff
@@ -7,82 +7,64 @@ import "./Chef.sol";
...
+    function stakeAdditionalMetric(uint256 metricAmount, uint256 newStartDate) public {
+        Staker storage stake = staker[msg.sender];
+        harvest();
+        uint256 principalMetric = stake.metricAmount;
         uint256 totalMetricStaked = metricAmount + principalMetric;
 
-        Staker memory stake = Staker({
-            stakeAddress: stakeAddress,
+        staker[msg.sender] = Staker({
+            stakeAddress: msg.sender,
```
jdaly13, line 59: Do we need thins any longer since we are doing the mapping?

src/contracts/StakingChef.sol
```diff
@@ -95,73 +77,72 @@ contract StakingChef is Chef {
...
 
-        emit Withdraw(msg.sender, stakeIndex, stake.metricAmount);
+        emit TransferPrincipal(msg.sender, stake, stake.metricAmount);
     }
 
-    function harvest(uint256 stakeIndex) internal {
-        Staker storage stake = _stakes[stakeIndex];
-
+    function harvest() internal {
+        Staker storage stake = staker[msg.sender];
```
jdaly13, line 105: if this is called internally I don't believe it has access to msg.sender - I think you would need to pass it in from whatever is invoking it

jdaly13: I think I maybe be wrong about this  - in sushiswap an internal function has access to msg.sender

src/contracts/StakingChef.sol
```diff
@@ -7,82 +7,64 @@ import "./Chef.sol";
...
+    // --------------------------------------------------------------------- staking functions
+    function stakeMetric(uint256 metricAmount, uint256 newStartDate) external nonDuplicated(msg.sender) {
+        if (areRewardsActive() && getTotalAllocationShares() > 0) {
             updateAccumulatedStakingRewards();
         }
-        Staker memory stake = Staker({
-            stakeAddress: newAddress,
-            metricAmount: metricAmount, 
+        staker[msg.sender] = Staker({
+            stakeAddress: msg.sender,
```
jdaly13, line 25: same as below not sure if this is needed anymore since it's in the mapping

src/contracts/StakingChef.sol
```diff
@@ -7,82 +7,64 @@ import "./Chef.sol";
...
-        uint256 metricAmount,
-        uint256 newStartDate
-    ) public {
-        harvest(stakeIndex);
-        uint256 principalMetric = _stakes[stakeIndex].metricAmount;
+    function stakeAdditionalMetric(uint256 metricAmount, uint256 newStartDate) public {
+        Staker storage stake = staker[msg.sender];
+        harvest();
+        uint256 principalMetric = stake.metricAmount;
         uint256 totalMetricStaked = metricAmount + principalMetric;
```
jdaly13, line 56: thinking we should stay consistent and use safeMath so .add

src/contracts/StakingChef.sol
```diff
@@ -95,73 +77,72 @@ contract StakingChef is Chef {
...
-    function withdrawPrincipal(uint256 stakeIndex) public {
-        Staker storage stake = _stakes[stakeIndex];
-
+    function withdrawPrincipal() public {
+        Staker storage stake = staker[msg.sender];
         require(stake.metricAmount != 0, "No Metric to withdraw");
-        require(stake.stakeAddress == _msgSender(), "Sender can not withdraw");
 
         SafeERC20.safeTransfer(IERC20(getMetricToken()), msg.sender, stake.metricAmount);
         stake.metricAmount = 0;
```
jdaly13, line 99: since this is storage I'm guessing you can use stake as a setter but if it was memory you'd have to use staker[msg.sender].metricAmount = 0; 
?

src/contracts/StakingChef.sol
```diff
@@ -95,73 +77,72 @@ contract StakingChef is Chef {
...
+    function viewPendingHarvest() public view returns (uint256) {
+        Staker storage stake = staker[msg.sender];
 
         return stake.metricAmount.mul(getLifetimeShareValue()).div(ACC_METRIC_PRECISION).sub(stake.rewardDebt);
     }
 
-    function viewPendingClaims(uint256 stakeIndex) public view returns (uint256) {
-        Staker storage stake = _stakes[stakeIndex];
+    function viewPendingClaims() public view returns (uint256) {
+        Staker storage stake = staker[msg.sender];
```
jdaly13, line 131: I'm thinking here you could use memory in place of storage? since its read only?

https://medium.com/coinmonks/ethereum-solidity-memory-vs-storage-which-to-use-in-local-functions-72b593c3703a
referencing getUsingMemory function in this article

src/contracts/StakingChef.sol
```diff
@@ -7,82 +7,64 @@ import "./Chef.sol";
...
-        addTotalAllocShares(_stakes[stakeIndex].metricAmount, metricAmount);
-        _stakes[stakeIndex].stakeAddress = stakeAddress;
-        _stakes[stakeIndex].metricAmount = metricAmount;
+        addTotalAllocShares(staker[msg.sender].metricAmount, metricAmount);
+        staker[msg.sender].metricAmount = metricAmount;
     }
 
-    function removeStaker(uint256 stakeIndex) external {
-        require(stakeIndex < _stakes.length, "Index more than stakes length");
+    function removeStaker() external {
```
jdaly13, line 44: I guess I'm having trouble distinguishing removeStaker from withdraw Principal in theory shouldn't they do the same thing?

jdaly13: I'm thinking both should be combined and be called unStakeMetric

src/contracts/StakingChef.sol
```diff
@@ -95,73 +77,72 @@ contract StakingChef is Chef {
...
+    function viewPendingClaims() public view returns (uint256) {
+        Staker storage stake = staker[msg.sender];
 
         return stake.claimable;
     }
 
-// --------------------------------------------------------------------- Structs
+    // --------------------------------------------------------------------- Structs
     struct Staker {
         address stakeAddress;
```
jdaly13, line 138: don't think this is needed

src/contracts/StakingChef.sol
gnomadic: it will work -  the other option is `tx.origin` which we do not want to use - usually ever.   It'll become more and more common for people to write their contracts to call into ours, and we should work as expected when that happens.

src/contracts/StakingChef.sol
gnomadic: honestly we can revisit all the External API's when ya build out the integration with the site right?  We don't have to get it perfect before it's used

gnomadic:
also why is package-lock.json in here but package.json isn't?  I dunno enough about node to know why would one change without the other,  I thought they were pretty tightly coupled

src/contracts/StakingChef.sol
gnomadic: I think you have to use storage because staker[msg.sender] is already in storage

PR #37, opened 2022-06-10
fix bug in deployment script

jdaly13:


deploy/00_deploy_protocol.js
```diff
@@ -18,7 +16,7 @@ module.exports = async (hre) => {
     log: true,
   });
 
-  if (chainId !== 31337 && hre.network.name !== "localhost") {
+  if (chainId !== "31337" && hre.network.name !== "localhost" && hre.network.name !== "hardhat") {
```
gnomadic, line 19: should these be `||` or not `&&`?  if not, cool, I don't quite follow the logic but I'm pretty sure we don't want to verify on local dev nodes right?

gnomadic (approved):

deploy/00_deploy_protocol.js
jdaly13: we want to make sure all three are falsy

PR #38, opened 2022-06-13
Investor chef

addyxvii:
This is the contract for investors.

### DIFF
--- a/src/contracts/Chef.sol
+++ b/src/contracts/Chef.sol
@@ -21,6 +21,7 @@ abstract contract Chef is Ownable {
     using SafeMath for uint256;
     uint256 private _metricPerBlock;
     uint256 public constant ACC_METRIC_PRECISION = 1e12;
+    AllocationGroup[] internal _allocations;
 
     bool private _rewardsActive;
     uint256 private _lastRewardBlock;
@@ -116,4 +117,14 @@ abstract contract Chef is Ownable {
 
     event Harvest(address harvester, uint256 agIndex, uint256 amount);
     event Withdraw(address withdrawer, uint256 agIndex, uint256 amount);
+
+    //------------------------------------------------------Support Functions
+
+    struct AllocationGroup {
+        address groupAddress;
+        uint256 shares;
+        bool autodistribute;
+        uint256 rewardDebt; // keeps track of how much the user is owed or has been credited already
+        uint256 claimable;
+    }
 }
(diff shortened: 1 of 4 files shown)


jdaly13:
Couldn't Investor Chef inherit from TopChef so InvestorChef is TopChef?
That way you could inherit all those methods and in TopChef you could mark any functions that aren't in Chef as virtual and if you need to override them in Investor Chef just mark as override?

gnomadic:
yeah I like that approach James - Addison do you see what he's saying?

addyxvii:
@gnomadic yep updates incoming

src/contracts/Chef.sol
```diff
@@ -116,4 +117,14 @@ abstract contract Chef is Ownable {
 
     event Harvest(address harvester, uint256 agIndex, uint256 amount);
     event Withdraw(address withdrawer, uint256 agIndex, uint256 amount);
+
+    //------------------------------------------------------Support Functions
+
+    struct AllocationGroup {
```
gnomadic, line 123: why did ya move this here?  StakingChef doesn't need it, right?

addyxvii: oh whoops that was an accident

src/contracts/InvestorChef.sol
```diff
@@ -0,0 +1,24 @@
...
+import "./TopChef.sol";
+
+contract InvestorChef is TopChef {
+    address public groupAddress;
+    uint256 public agIndex;
+    uint256 public shares;
+    bool public newAutoDistribute;
+
+    constructor(address metricTokenAddress) TopChef(metricTokenAddress) {
+        setMetricToken(metricTokenAddress);
```
gnomadic, line 14: wait doesn't this call every function from the constructor?

addyxvii: Yeah which I thought was odd but took your message too literally ha ha. What is the solidity pattern? 
```
function methodContainer {
          parentMethod1(),
          parentMethod2(),
    }
 ```
 or
 ```
 function method1 {
          parentMethod1()
     }
 function method2 {
           parentMethod2()
    }
 ```

gnomadic: no don't need anything!  it should just inherit em

`contract InvestorChef is TopChef {
    constructor(address metricTokenAddress) TopChef(metricTokenAddress) {
}
}
`

gnomadic (approved):

PR #39, opened 2022-06-17
bug fixes

jdaly13:
bug fix for viewPendingHarvest - create a new method that replicates adding to lifeTimeShareValue so it's more of an estimate - change rewardsDebt for updated allocation groups

### DIFF
--- a/src/contracts/Chef.sol
+++ b/src/contracts/Chef.sol
@@ -54,6 +54,13 @@ abstract contract Chef is Ownable {
         _lifetimeShareValue = _lifetimeShareValue.add(accumulatedMetricDividedByShares(accumulatedWithMetricPrecision));
     }
 
+    function getLifeTimeShareValueEstimate() public view virtual returns (uint256) {
+        uint256 accumulated = getAccumulated();
+        uint256 accumulatedWithMetricPrecision = getAcculatedWithmetricPrecision(accumulated);
+        uint256 lifetimesharevalue = getLifetimeShareValue();
+        return lifetimesharevalue.add(accumulatedMetricDividedByShares(accumulatedWithMetricPrecision));
+    }
+
     function addTotalAllocShares(uint256 shares) internal virtual {
         _totalAllocShares = _totalAllocShares.add(shares);
     }
--- a/src/contracts/TopChef.sol
+++ b/src/contracts/TopChef.sol
@@ -1,4 +1,3 @@
-
 //SPDX-License-Identifier: Unlicense
 pragma solidity ^0.8.0;
 
@@ -22,7 +21,7 @@ contract TopChef is Chef {
         address newAddress,
         uint256 newShares,
         bool newAutoDistribute
-    ) external onlyOwner() nonDuplicated(newAddress) {
+    ) external onlyOwner nonDuplicated(newAddress) {
         if (areRewardsActive() && getTotalAllocationShares() > 0) {
             updateAccumulatedAllocations();
         }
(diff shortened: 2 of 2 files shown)


src/contracts/TopChef.sol
```diff
@@ -38,23 +37,24 @@ contract TopChef is Chef {
...
-    ) public onlyOwner() {
+    ) public onlyOwner {
         if (areRewardsActive() && getTotalAllocationShares() > 0) {
             updateAccumulatedAllocations();
         }
         addTotalAllocShares(_allocations[agIndex].shares, shares);
         _allocations[agIndex].groupAddress = groupAddress;
         _allocations[agIndex].shares = shares;
         _allocations[agIndex].autodistribute = newAutoDistribute;
+        _allocations[agIndex].rewardDebt = shares.mul(getLifetimeShareValue()).div(ACC_METRIC_PRECISION).sub(_allocations[agIndex].rewardDebt);
```
gnomadic, line 54: actually wait - if it's a call to update, they already have existing shares and are earning rewards right?  so would we want to zero it out?

src/contracts/Chef.sol
```diff
@@ -54,6 +54,13 @@ abstract contract Chef is Ownable {
         _lifetimeShareValue = _lifetimeShareValue.add(accumulatedMetricDividedByShares(accumulatedWithMetricPrecision));
     }
 
+    function getLifeTimeShareValueEstimate() public view virtual returns (uint256) {
```
gnomadic, line 57: should we make `getLifetimeShareValue()` private to avoid a confusing public API?

src/contracts/TopChef.sol
jdaly13: when we are updating I thought we were essentially resetting?
Anytime we add an allocation group we do
```newShares.mul(getLifetimeShareValue()).div(ACC_METRIC_PRECISION)```
so I thought we should essentially do the same minus their existing reward debt which would be like, in my view you are just resetting it back to what it would be when you do an additional add - but I might be wrong about this
@addyxvii do you have any thoughts about this?

src/contracts/Chef.sol
jdaly13: sure I can set it to internal - thinking of renaming it to getLastRecordedLifeTimeShareValue because essentially that's what it is - you need to call `setLifetimeShareValue` before you get the latest value - it's essentially a getter after a recent setter - but the value is only up to date after calling `setLifetimeShareValue`

jdaly13: `getLifetimeShareValue` is now internal @gnomadic 
at @addyxvii let me know if you have thoughts about rename as well or is it making it to explicit

addyxvii: I think it's a good name, don't think there's a better way to indicate that it's an imprecise value

src/contracts/TopChef.sol
jdaly13: _allocations[agIndex].rewardDebt = shares.mul(getLifetimeShareValue()).div(ACC_METRIC_PRECISION).sub(_allocations[agIndex].rewardDebt);
which does the same thing as harvest I believe without transferring or the claiming of tokens
for example
uint256 claimable = group.shares.mul(getLifetimeShareValue()).div(ACC_METRIC_PRECISION).sub(group.rewardDebt);
group.rewardDebt = claimable;
that statement in harvest (right above) essentially assigns the same value to rewardDebt as the new statement I created in updateAllocationGroup (first code block)
At least that's what I perceive that is why I did it that way

addyxvii (approved):

PR #43, opened 2022-06-23
xMETRIC token logic & tests

thermohaline:
xMETRIC

### DIFF
--- a/src/contracts/xMETRIC.sol
+++ b/src/contracts/xMETRIC.sol
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+/// @notice We use solmate as it is more gas efficient thab OZ
+/// @notice https://github.com/Rari-Capital/solmate/blob/main/src/tokens/ERC20.sol
+/// @notice  https://github.com/Rari-Capital/solmate/blob/main/src/auth/Owned.sol
+
+import {ERC20} from "solmate/tokens/ERC20.sol";
+import {Owned} from "solmate/auth/owned.sol";
+
+contract xMETRIC is ERC20("xMETRIC", "xMETRIC", 18), Owned {
+    constructor(address _owner, uint256 initialSupply) Owned(_owner) {
+        _mint(msg.sender, initialSupply);
+    }
+
+    //------------------------------------------------------Overrides
+    function transfer(address to, uint256 amount)
+        public
+        override
+        transactor
+        returns (bool)
+    {
+        balanceOf[msg.sender] -= amount;
+
+        // Cannot overflow because the sum of all user
+        // balances can't exceed the max uint256 value.
+        unchecked {
+            balanceOf[to] += amount;
+        }
+
+        emit Transfer(msg.sender, to, amount);
+
+        return true;
+    }
+
+    function transferFrom(
+        address from,
+        address to,
+        uint256 amount
+    ) public override transactor returns (bool) {
+        uint256 allowed = allowance[from][msg.sender]; // Saves gas for limited approvals.
+
+        if (allowed != type(uint256).max)
+            allowance[from][msg.sender] = allowed - amount;
+
+        balanceOf[from] -= amount;
+
+        // Cannot overflow because the sum of all user
+        // balances can't exceed the max uint256 value.
+        unchecked {
+            balanceOf[to] += amount;
+        }
+
+        emit Transfer(from, to, amount);
+
+        return true;
+    }
+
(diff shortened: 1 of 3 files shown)


src/contracts/xMETRIC.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+/// @notice We use solmate as it is more gas efficient thab OZ
+/// @notice https://github.com/Rari-Capital/solmate/blob/main/src/tokens/ERC20.sol
+/// @notice  https://github.com/Rari-Capital/solmate/blob/main/src/auth/Owned.sol
+
+import {ERC20} from "solmate/tokens/ERC20.sol";
+import {Owned} from "solmate/auth/owned.sol";
+
```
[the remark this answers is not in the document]
jdaly13: I would be in favor of keeping it consistent throughout our packages/contracts - I'm not really aware of solmate and we have been utilizing open zeppelin - so unless there's a reason to use solmate (I'm willing to be persuaded) I would think using the open zeppelin versions for these contract would be prudent - we can certainly meet to discuss if @thermohaline or anyone else feels strongly of using one over the other

src/contracts/xMETRIC.sol
```diff
@@ -0,0 +1,83 @@
...
+        public
+        override
+        transactor
+        returns (bool)
+    {
+        balanceOf[msg.sender] -= amount;
+
+        // Cannot overflow because the sum of all user
+        // balances can't exceed the max uint256 value.
+        unchecked {
```
jdaly13, line 27: good point - I was reading about unchecked and thinking it could save on some gas costs and being able to utilize when it's safe in other contracts as well

src/contracts/xMETRIC.sol
```diff
@@ -0,0 +1,83 @@
...
+        return true;
+    }
+
+    //------------------------------------------------------Setters
+    function addTransactor(address _transactor) public onlyOwner {
+        canTransact[_transactor] = true;
+    }
+
+    //------------------------------------------------------Getters
+    function getOwner() public view returns (address) {
```
jdaly13, line 65: is there a benefit in having this function?  if so couldn't you just make it a public variable so it automatically has a getter?

jdaly13 (commented):
Looks good to me - but I think we should just have a discussion regarding the use of solmate vs open zeppelin before merging in - also never written tests in solidity so I can't comment much on that other than I should start learning it

src/contracts/xMETRIC.sol
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    function isTransactor(address _addr) public view returns (bool) {
+        return canTransact[_addr];
+    }
+
+    //------------------------------------------------------Support Functions
+    mapping(address => bool) public canTransact;
+
+    error AddressCannotTransact();
```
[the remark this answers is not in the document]
gnomadic: I've also heard it costs less gas than a `require`

src/contracts/xMETRIC.sol
thermohaline: Agreed

src/contracts/xMETRIC.sol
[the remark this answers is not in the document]
thermohaline: Yep error messages should be easier to handle on the front end as well, and they indeed are cheaper than require 👍

src/contracts/xMETRIC.sol
thermohaline: Yeah exactly, was having some trouble earlier -- will resolve.

thermohaline (commented):
This should do the trick. Have:
- Removed getOwner()
- Switch to OZ
- Updated tests to support OZ

Screenshot 2022-06-23 at 19 36 48

Thanks!

addyxvii (approved):
I do think we should discuss solmate vs OZ as a team like @jdaly13 said, but this looks good to me so I'm approving

src/test/xMETRIC.t.sol
```diff
@@ -0,0 +1,147 @@
...
+        metricToken = new xMETRIC(1000000000 * 10**18);
+        vm.stopPrank();
+    }
+
+    function testExample() public {
+        assertTrue(true);
+    }
+
+    function test_GetTokenName() public {
+        console.log("Test ERC20 initialization");
```
jdaly13, line 29: Does logging work in foundry?  do we need it - it's a test so I don't care regardless, normally this would go in a describe or it statement when testing with javascript but not sure how testing works here

PR #47, opened 2022-06-25
fixing tests and standardizing on 0.8.13

gnomadic:


src/contracts/MetricToken.sol
```diff
@@ -1,11 +1,11 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
 /// @custom:security-contact contracts@metricsdao.com
 contract MetricToken is ERC20 {
-    constructor(address _vestingContractAddress) ERC20("METRIC", "METRIC") {
```
gnomadic, line 8: I did this because as James pointed out we will deploy Metric before we deploy the vesting contract

test/chef-test.js
```diff
@@ -7,29 +7,30 @@ describe("Allocator Contract", function () {
...
 
+  let origin;
   let allocationGroup1;
   let allocationGroup2;
-  let vestingContract;
+  let addrs;
 
   beforeEach(async function () {
-    [allocationGroup1, allocationGroup2, vestingContract] = await ethers.getSigners();
+    [origin, allocationGroup1, allocationGroup2, ...addrs] = await ethers.getSigners();
```
gnomadic, line 16: new convention proposal:  let's do `origin` as the first account because they are the `owner` and will usually have the minted supply.  We shouldn't use this account for "user" tests because they aren't a user.

thermohaline (approved):
Looks good

jdaly13 (approved):

PR #51, opened 2022-06-28
refactor claim and harvest based metrics dao discussions

jdaly13:
based on discussions - claim should always call harvest so user gets latest rewards - also harvest and harvest all should be callable by any one


src/contracts/TopChef.sol
```diff
@@ -115,47 +119,32 @@ contract TopChef is Chef {
...
-        if ((viewPendingHarvest(agIndex) == 0)) revert NoRewardsToClaim();
         AllocationGroup storage group = _allocations[agIndex];
         // TODO do we want a backup in case a group looses access to their wallet
-        if (!(group.groupAddress == _msgSender() || _msgSender() == owner())) revert SenderNotGroupOrOwner();
 
         // Effects
         updateAccumulatedAllocations();
-
-        uint256 toClaim = (group.shares * getLifetimeShareValue()) / ACC_METRIC_PRECISION - group.rewardDebt;
+        uint256 toClaim = ((group.shares * (getLifetimeShareValue())) / ACC_METRIC_PRECISION) - group.rewardDebt;
```
jdaly13, line 130: PEMDAS

src/contracts/TopChef.sol
```diff
@@ -115,47 +119,32 @@ contract TopChef is Chef {
...
 
         group.rewardDebt = group.rewardDebt + toClaim;
+        uint256 totalClaimable = group.claimable + toClaim;
+        group.claimable = totalClaimable;
```
jdaly13, line 135: removed autoDistribute - as it's really not needed anymore - right not the rest of it is dead code - we should create a ticket to remove all autoDistribute

thermohaline (approved):
Looks good

PR #53, opened 2022-06-29
Update TopChef.sol modifiers & add TopChef foundry tests

thermohaline:
- TopChef coverage for foundry
- Cleaning up the code in relation to the rewards toggle


src/contracts/TopChef.sol
```diff
@@ -57,10 +56,9 @@ contract TopChef is Chef {
         _allocations[agIndex].autodistribute = newAutoDistribute;
     }
 
-    function removeAllocationGroup(uint256 agIndex) external onlyOwner {
+    function removeAllocationGroup(uint256 agIndex) external activeRewards onlyOwner {
         // Checks
         if (!(agIndex < _allocations.length)) revert IndexDoesNotMatchAllocation();
```
addyxvii, line 61: Suggest changing this to ```if (agIndex > _allocations.length) revert IndexDoesNotMatchAllocation();```

jdaly13: also I think it should be ` if (agIndex >= _allocations.length) revert IndexDoesNotMatchAllocation();`

jdaly13 (changes requested):
https://github.com/MetricsDAO/core-evm-contracts/pull/53/files#r910212943

gnomadic (approved):

PR #56, opened 2022-07-05
adding todos after product discussion

gnomadic:


thermohaline:
- Added minting on transfer
- Added burn functionality
- Removed initial supply

src/contracts/xMETRIC.sol
```diff
@@ -21,13 +21,22 @@ contract xMETRIC is ERC20("xMETRIC", "xMETRIC"), Ownable {
         address to,
         uint256 amount
     ) public override transactor returns (bool) {
-        address spender = _msgSender();
-        _spendAllowance(from, spender, amount);
-        _transfer(from, to, amount);
+        _spendAllowance(from, msg.sender, amount);
```
jdaly13, line 24: I guess this slightly confuses me now - so to clarify
first we need addTransactor needs to be called before calling this which is essentially whitelisting an address - we check to see if msg.sender is whitelisted first
then we check to see if the from address, owner of current token has enough token which is passed in the amount param in their wallet.
then this contract just mints/creates new xMetric "amount" param and sends it to the "to" param address. 
From there we remove/burn the "amount" param of tokens in the from address?
Basically anyone who is a transactor can call this on behalf of someone else which seems odd
Also is there an easier way to do this?  Maybe not but took a minute for me to grasp

jdaly13:
Also have we checked out these extensions on ERC20
https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#extensions
Mintable, Burnable, Pausable, Capped?

src/contracts/xMETRIC.sol
thermohaline: My understanding was this:
- We create a monopoly money token (one without an initial or capped supply) that only certain addresses are able to transact (the transactors).
- We add an address as transactor, this address is now allowed to access both transfer and transferFrom functionality.
- As discussed, we replace _transfer with _mint in order to send around arbitrary amounts of tokens.

The use-cases for the function are as followed:
- A transactor (which in my understanding will be accounts controlled/permitted by us) can "transfer" tokens to any user. This directly mints the tokens to the user. We do this because we don't initially distribute supply to ourselves. The receiver of the tokens we could call Alice. 
- Alice can't transfer these tokens as she is not a transactor. However, we want to also be able to test masterchef functionality with these xMetric tokens for instance. So what Alice can do is approve a masterchef contract, where the master contract is added as transactor by the owner. The masterchef contract can then access Alice's funds by using TransferFrom. Here we mimic a transfer by "transfering" the tokens from Alice to the masterchef contract for instance. This is done by decreasing Alice's token balance (by burning) while also decreasing the allowance that Alice has set for the masterchef contract, and increasing the masterchef balance (by minting). This lets us verify the accounting and mimic the accounting.

Additionally, we added burn functionality where any user can burn their own tokens, and the owner can burn anyones tokens.

And it should only be possible for a transactor to spend balance of a user that first approved the transactor to spend set balance, otherwise it will still revert with ← "ERC20: insufficient allowance"

Screenshot 2022-07-06 at 18 49 21
.

jdaly13:
@thermohaline @gnomadic - I think we discussed adding Pausable functionality `ERC20Pausable` 
so we create a function that only owner could call like 
`function pause() onlyOwner {
 _pause()
 }`
 
 also do we want to combine the functionality of approve and transferFrom for participants like Alice who can receive tokens, burn tokens, but can't not transfer tokens directly but she can approve another address who is a "transactor" and then call `transferFrom`

src/contracts/xMETRIC.sol
```diff
@@ -5,14 +5,14 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
 contract xMETRIC is ERC20("xMETRIC", "xMETRIC"), Ownable {
```
jdaly13, line 7: `contract xMETRIC is ERC20("xMETRIC", "xMETRIC"), ERC20Pausable, Ownable  {`

And in the contract itself add two functions
`    function pause() public onlyOwner {
        _pause()
    }`
    
    ` function unPause() public onlyOwner {
        _unpause()
    }`

gnomadic: since we have a totally custom transfer function I don't think we can/should rely on openzeppelin for this one - might not work out of the box

gnomadic:
> @thermohaline @gnomadic - I think we discussed adding Pausable functionality `ERC20Pausable` so we create a function that only owner could call like `function pause() onlyOwner { _pause() }`
> 
> also do we want to combine the functionality of approve and transferFrom for participants like Alice who can receive tokens, burn tokens, but can not transfer tokens directly but she can approve another address who is a "transactor" and then call `transferFrom`

I don't think we should combine them because in a "real" erc20 they are not combined - even though this is funky, let's try and modify the protocol as little as possible to support it.. does that sound right to ya?

jdaly13:
> > @thermohaline @gnomadic - I think we discussed adding Pausable functionality `ERC20Pausable` so we create a function that only owner could call like `function pause() onlyOwner { _pause() }`
> > also do we want to combine the functionality of approve and transferFrom for participants like Alice who can receive tokens, burn tokens, but can not transfer tokens directly but she can approve another address who is a "transactor" and then call `transferFrom`
> 
> I don't think we should combine them because in a "real" erc20 they are not combined - even though this is funky, let's try and modify the protocol as little as possible to support it.. does that sound right to ya?

Sounds good - I was just getting caught up in that one scenario I guess, it's covered in the test which is the most important part that approve needs to be called first

src/contracts/xMETRIC.sol
jdaly13: this functionality is called in  function _beforeTokenTransfer(
which is called before _mint

jdaly13 (approved):

PR #57, opened 2022-07-06
admin can dq question

addyxvii:
Admin can mark a question as bad


src/contracts/Protocol/QuestionStateController.sol
```diff
@@ -29,6 +29,10 @@ contract QuestionStateController is IQuestionStateController, Ownable {
         votes[questionId].totalVoteCount = votes[questionId].totalVoteCount + amount;
     }
 
+    function setBadState(uint256 questionId) public onlyOwner {
```
jdaly13, line 32: I guess it doesn't matter because it's onlyOwner but should this be internal

jdaly13 (approved):
LGTM - just had a couple of questions but nothing blocking

PR #59, opened 2022-07-07
add pause and unpause functionality rename contract for linter

jdaly13:
adding pause and unpause functionality 
no tests for this - could add later - we don't currently have JS tests for this and could work on that

### DIFF
--- a/src/contracts/xMETRIC.sol
+++ b/src/contracts/xMETRIC.sol
@@ -4,7 +4,7 @@ pragma solidity 0.8.13;
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
-contract xMETRIC is ERC20("xMETRIC", "xMETRIC"), Ownable {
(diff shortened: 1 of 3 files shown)


jdaly13:
every time you call _burn or _mint - it would check if the contract is paused or unpaused
it does this in _beforeTokenTransfer

src/contracts/xMETRIC.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.13;
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
-contract xMETRIC is ERC20("xMETRIC", "xMETRIC"), Ownable {
+contract XmETRIC is ERC20("xMETRIC", "xMETRIC"), ERC20Pausable, Ownable {
```
jdaly13, line 7: Our Linter and I believe it's a common pattern to always name your contract with an Uppercase

gnomadic (approved):
AH I don't like the Xmetric change haha BUT let's keep our linter happy.  @thermohaline FYI. ^^

PR #73, opened 2022-07-07
General mdao refactoring

thermohaline:
- General refactoring of some mdao logic

### DIFF
--- a/src/contracts/Protocol/BountyQuestion.sol
+++ b/src/contracts/Protocol/BountyQuestion.sol
@@ -20,7 +20,7 @@ import "./onlyApi.sol";
 // philosphy -> start out gated and the protocol can evolve in the same way the dao does
 
 /// @custom:security-contact contracts@metricsdao.xyz
-contract BountyQuestion is ERC721, ERC721Enumerable, ERC721URIStorage, ERC721Burnable, Ownable, onlyApi {
+contract BountyQuestion is ERC721, ERC721Enumerable, ERC721URIStorage, ERC721Burnable, Ownable, OnlyApi {
     using Counters for Counters.Counter;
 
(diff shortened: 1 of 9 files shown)


src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -29,7 +29,7 @@ contract BountyQuestion is ERC721, ERC721Enumerable, ERC721URIStorage, ERC721Bur
     constructor() ERC721("MetricsDAO Question", "MDQ") {}
 
     // working standard metadata format:  Title, Description, Program
-    function safeMint(address to, string memory uri) public onlyOwner returns (uint256) {
+    function safeMint(address to, string calldata uri) public returns (uint256) {
```
gnomadic, line 32: I think we want this to be `onlyAPI` because if someone calls `safeMint` but doesn't call initialize on the other controllers state will be out of whack

src/contracts/Protocol/ClaimController.sol
```diff
@@ -36,6 +38,26 @@ contract ClaimController is Ownable, IClaimController {
...
+    error ClaimLimitReached();
+    error NeedClaimToAnswer();
+
+    // Api stuff
+    function setQuestionApi(address _questionApi) public onlyOwner {
+        questionApi = _questionApi;
+    }
+
+    // ------------------------------- Modifier
+    error NotTheApi();
```
gnomadic, line 55: Maybe this modifier should be in it's own solidity file so we can import it into every controller?

thermohaline: Yeah sounds good

gnomadic (approved):

PR #79, opened 2022-07-11
Add possibility to update question creation costs & test

thermohaline:
Added possibility to update question creation costs & test

### DIFF
--- a/src/contracts/Protocol/QuestionAPI.sol
+++ b/src/contracts/Protocol/QuestionAPI.sol
@@ -84,8 +84,16 @@ contract QuestionAPI is Ownable {
     function answerQuestion(uint256 questionId, string calldata answerURL) public {
         _claimController.answer(questionId, answerURL);
     }
-    
-     function disqualifyQuestion(uint256 questionId) public onlyOwner {
+
+    /**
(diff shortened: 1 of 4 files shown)


src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -84,8 +84,16 @@ contract QuestionAPI is Ownable {
...
         _claimController.answer(questionId, answerURL);
     }
-    
-     function disqualifyQuestion(uint256 questionId) public onlyOwner {
+
+    /**
+     * @notice Changes the cost of creating a question
+     * @param cost The new cost of creating a question
+     */
+    function setCreateCost(uint256 cost) public onlyOwner {
```
gnomadic, line 92: I don't think we need this on the QuestionAPI, I think it can just be on the _costContoller... right?  Like this function will be called very infrequently and only by admins

thermohaline: Yep you're right. Had a mental error thinking we'd have to transfer ownership which is not the case. Updated!

gnomadic (approved):

PR #81, opened 2022-07-11
adding script to set transactor

gnomadic:


scripts/xMetric/addTransactor.ts
```diff
@@ -0,0 +1,26 @@
+import { ethers } from "hardhat";
+import fs from "fs";
+
+async function main() {
+  const newTrasactor = "";
```
jdaly13, line 5: can this be blank?

scripts/xMetric/addTransactor.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const network = process.env.HARDHAT_NETWORK;
+  console.log("adding transactor on: ", network);
+
+  const data = await fs.readFileSync("./deployments/" + network + "/Xmetric.json", "utf8");
+  const object = JSON.parse(data);
+  const address = object.address;
+
+  const MyContract = await ethers.getContractFactory("Xmetric");
```
jdaly13, line 14: Guess I'm confused from the hardhat documentation

> The Contract s and ContractFactory  returned by these helpers are connected to the first [signer](https://docs.ethers.io/v5/single-page/#/v5/api/signer/) returned by getSigners by default.

How does it know which signer it is - I'm assuming the owner of the contract should be the signer but where is the relationship established?

scripts/xMetric/addTransactor.ts
gnomadic: yeah the intention is that when we run the script against mainnet or whatever we would manually add the person we want to add as a transactor here

scripts/xMetric/addTransactor.ts
```diff
@@ -0,0 +1,26 @@
...
+
+async function main() {
+  const newTrasactor = "";
+
+  const network = process.env.HARDHAT_NETWORK;
+  console.log("adding transactor on: ", network);
+
+  const data = await fs.readFileSync("./deployments/" + network + "/Xmetric.json", "utf8");
+  const object = JSON.parse(data);
+  const address = object.address;
```
jdaly13, line 12: the problem I see with this is if we are using forge to deploy I'm not sure exaclty how it format is compared to hardhat-deploy which adds an address

scripts/xMetric/addTransactor.ts
gnomadic: it pulls from `hardhat.config.ts` - like on ropsten it would use this account: https://github.com/MetricsDAO/core-evm-contracts/blob/main/hardhat.config.ts#L32

test/Xmetric-test.js
```diff
@@ -52,7 +52,6 @@ describe("xMETRIC", async function () {
     // deploy Factory
     const factoryContract = await ethers.getContractFactory("QuestionAPI");
     questionAPI = await factoryContract.deploy(
-      xMetric.address,
```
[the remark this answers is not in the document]
gnomadic: I'm under the impression that those json files are created by `npx hardhat deploy` and I used the conventions of the existing deployments

scripts/xMetric/addTransactor.ts
gnomadic: yeah let's use hardhat to deploy for now - I really find a ton of value in these json files

scripts/xMetric/addTransactor.ts
jdaly13: I don't see anything for polygon on hardhat config do you want to add and just use your .env file?

gnomadic: yeah I was gonna do it live haha but good point we can prep.

scripts/xMetric/addTransactor.ts
jdaly13: right but in this context - if there's nothing there will the transaction revert? couldn't you just use an .env variable?

gnomadic: oh it'll fail when it tries to convert an empty string into a an `address` -> but yeah -> we aren't adding any transactors right now..  probably not till next week or later in this one?  We can't add a transactor without knowing their address anyways, so this script is just making it easy to plug in their address when we're ready for it

jdaly13: ah okay - was slightly confused - If you wanted to - you could just call pause for now and then unpause it when you're ready to do transactions but at least with pause you'll now its working properly

gnomadic: I'll add another two scripts to handle pause and unpause but I like having one function per script

jdaly13 (approved):

PR #93, opened 2022-07-12
_msgSender and cleanup

thermohaline:


src/contracts/Xmetric.sol
```diff
@@ -5,9 +5,9 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Pausable.sol";
 
-contract Xmetric is ERC20("xMetric", "xMETRIC"), ERC20Pausable, Ownable {
+contract Xmetric is ERC20("Xmetric", "xMETRIC"), ERC20Pausable, Ownable {
```
gnomadic, line 8: it needs to be xMetric!  is this our linter?

thermohaline: I updated this because the tests were failing, it was expecting Xmetric. We deployed Xmetric to polygon as well

jdaly13: so this is deployed to polygon let's keep it 
```contract Xmetric is ERC20("Xmetric", "xMETRIC"), ERC20Pausable, Ownable {```
these two tests need to be updated I think they got mixed
```

    function test_GetTokenName() public {
        console.log("Test ERC20 initialization");
        assertEq(metricToken.name(), "xMetric");
    }

    function test_GetTokenSymbol() public {
        assertEq(metricToken.symbol(), "xMETRIC");
    }
```

jdaly13: @thermohaline  Please pull from latest - this should be fixed so you don't need to make any changes

jdaly13:
Please update merge conflict and also can we not delete teh MetricToken test and just fix it? - just needs to be uncommented out

jdaly13:
Also is there a reason to delete the Chef Test file?

jdaly13 (changes requested):
Just want to clarify before we merge in why we are deleting these two test files?

thermohaline:
> Just want to clarify before we merge in why we are deleting these two test files?

The tests are covered more extensively in the other test files, the files only contain 2 very basic tests

PR #95, opened 2022-07-13
update JS tests and bount question

jdaly13:
WIP

### DIFF
--- a/src/contracts/Protocol/BountyQuestion.sol
+++ b/src/contracts/Protocol/BountyQuestion.sol
@@ -24,7 +24,9 @@ contract BountyQuestion is ERC721, ERC721Enumerable, ERC721URIStorage, ERC721Bur
     using Counters for Counters.Counter;
 
     Counters.Counter private _tokenIdCounter;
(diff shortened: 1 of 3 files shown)


thermohaline (approved):

src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -59,4 +62,8 @@ contract BountyQuestion is ERC721, ERC721Enumerable, ERC721URIStorage, ERC721Bur
     function supportsInterface(bytes4 interfaceId) public view override(ERC721, ERC721Enumerable) returns (bool) {
         return super.supportsInterface(interfaceId);
     }
+
+    function getAuthor(address user) public view returns (uint256[] memory) {
```
gnomadic, line 66: since it's public you don't actually need this!  `authors(address user)` will be auto-generated by solidity

jdaly13: I can't find the correct format for the second argument regardless and I think it's just easier to read as well but curious to know what the format would be but not going down that rabbit hole at this time

PR #97, opened 2022-07-13
Add createChallenge and implement first NFT Locked function

thermohaline:
- Added a first function (createChallenge) that is NFTLocked
- Created some tests to verify the behavior

### DIFF
--- a/src/contracts/Protocol/Extra/MockAuthNFT.sol
+++ b/src/contracts/Protocol/Extra/MockAuthNFT.sol
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.13;
+
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+
+contract NFT is ERC721 {
+    uint256 public currentTokenId;
+
+    constructor(string memory _name, string memory _symbol) ERC721(_name, _symbol) {}
+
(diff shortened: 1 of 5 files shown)


src/contracts/Protocol/modifiers/NFTLocked.sol
```diff
@@ -5,6 +5,8 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/interfaces/IERC721.sol";
 
 abstract contract NFTLocked is Ownable {
+    bytes32 public constant PROGRAM_MANAGER_ROLE = keccak256("PROGRAM_MANAGER_ROLE");
```
gnomadic, line 8: I wonder if there's a better place to stash these?  for now this will work - but as we add a ton of em, maybe in the file that uses the role?  or maybe in the nft that defines the role?

src/contracts/Protocol/Extra/MockAuthNFT.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: UNLICENSED
```
gnomadic, line 1: can ya start setting MIT license?


```suggestion
// SPDX-License-Identifier: MIT
```

gnomadic (approved):

PR #105, opened 2022-07-14
add deployment scripts and local deployments

jdaly13:
new deployment script for question API


deploy/01_deploy_questionapi.js
```diff
@@ -0,0 +1,92 @@
+module.exports = async (hre) => {
+  const { getNamedAccounts, deployments, getChainId } = hre;
+  const { deploy } = deployments;
+  const { deployer } = await getNamedAccounts();
+  const chainId = await getChainId();
+
+  const whichMetric = process.env.metric === "metric" ? "MetricToken" : "Xmetric";
```
gnomadic, line 7: this is wondeful

deploy/02_deploy_protocol.js
```diff
@@ -0,0 +1,56 @@
...
+    log: true,
+  });
+
+  const stakingChef = await deploy("StakingChef", {
+    from: deployer,
+    args: [metricToken.address],
+    log: true,
+  });
+
+  if (chainId !== "31337" && hre.network.name !== "localhost" && hre.network.name !== "hardhat") {
```
jdaly13, line 24: could use a promise.all as this seems overly redundant but oh well

PR #107, opened 2022-07-15
calling claim when removing an AG and fixing tests

gnomadic:


src/contracts/TopChef.sol
```diff
@@ -15,11 +15,7 @@ contract TopChef is Chef {
...
     //------------------------------------------------------Manage Allocation Groups
 
-    function addAllocationGroup(
-        address newAddress,
-        uint256 newShares,
-        bool newAutoDistribute
-    ) external onlyOwner nonDuplicated(newAddress) {
+    function addAllocationGroup(address newAddress, uint256 newShares) external onlyOwner nonDuplicated(newAddress) {
         // Checks
         if (!(newShares > 0)) revert SharesNotGreaterThanZero();
```
addyxvii, line 20: Should we just do 
```if (newShares <= 0) revert SharesNotGreaterThanZero();```

addyxvii (approved):

src/contracts/TopChef.sol
gnomadic: yes good catch!

PR #109, opened 2022-07-18
Fixing remaining audit items

gnomadic:


src/contracts/Chef.sol
```diff
@@ -4,20 +4,9 @@ pragma solidity 0.8.13;
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "./MetricToken.sol";
 
-// POSSIBLE FUTURE ITERATIONS
```
jdaly13, line 7: Are we not doing these anymore or just getting rid of TODO's should we discuss?

jdaly13 (approved):
Looks fine to me - question could any of these be converted to modifiers would that make it cleaner at all or more modular more of a question but I think this is fine

PR #112, opened 2022-07-19
62 lock metric from api

addyxvii:
Start for Locking Metric functionality


src/contracts/Vault.sol
```diff
@@ -0,0 +1,103 @@
...
+        return IERC20(_metric).balanceOf(address(this));
+    }
+
+    event Withdraw(address withdrawer, uint256 amount);
+
+    struct Items {
+        address withdrawer;
+        uint256 amount;
+        uint256 unlockTimestamp;
+        bool withdrawn;
```
gnomadic, line 98: should this be an enum?  Are these states mutually exclusive?  it might be a pain in the ass to micromanage all of em together

src/contracts/Vault.sol
```diff
@@ -0,0 +1,103 @@
...
+            lockedMetric[_id].amount
+        );
+
+        emit Withdraw(msg.sender, lockedMetric[_id].amount);
+        lockedMetric[_id].metric.safeTransfer(msg.sender, lockedMetric[_id].amount);
+    }
+
+    //TODO: Double check math
+    function slashMetric(uint256 _id) external onlyOwner {
+        require(lockedMetric[_id].slashed = false);
```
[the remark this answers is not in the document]
addyxvii: Yep, I had forgotten updating now

src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -19,12 +20,14 @@ contract QuestionAPI is Ownable, NFTLocked {
...
         address claimController,
-        address costController
+        address costController,
+        address vault
     ) {
         _question = BountyQuestion(bountyQuestion);
         _questionStateController = IQuestionStateController(questionStateController);
         _claimController = IClaimController(claimController);
         _costController = IActionCostController(costController);
+        _vault = Vault(vault);
```
gnomadic, line 30: should this be on the `CostController` instead of the API?  not sure which is better

src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -50,6 +53,9 @@ contract QuestionAPI is Ownable, NFTLocked {
         _questionStateController.initializeQuestion(questionId);
         _claimController.initializeQuestion(questionId, claimLimit);
 
+        // Lock Metric for the question
+        _vault.lockMetric(questionId, _msg.sender(), 1e18);
```
gnomadic, line 57: either way you approach the comment above - the cost here (1e18) should come from the costController

src/contracts/Protocol/QuestionAPI.sol
addyxvii: I was going back and forth but I think the cost controller might make more sense

thermohaline (approved):
🤝

PR #118, opened 2022-07-21
Removing 721 from BountyQuestion

gnomadic:


src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -9,63 +9,40 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/utils/Counters.sol";
 import "./modifiers/OnlyAPI.sol";
```
jdaly13, line 10: Shouldn't we be removing imports?

src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -9,63 +9,40 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
-
-    function tokenURI(uint256 tokenId) public view override(ERC721, ERC721URIStorage) returns (string memory) {
-        return super.tokenURI(tokenId);
-    }
-
-    function supportsInterface(bytes4 interfaceId) public view override(ERC721, ERC721Enumerable) returns (bool) {
-        return super.supportsInterface(interfaceId);
+        questions[tokenId] = QuestionData({tokenId: tokenId, url: uri});
+        // QuestionData memory _question = QuestionData({tokenId: tokenId, url: uri});
+        authors[author].push(tokenId);
```
jdaly13, line 34: I think I get it but if I query by author I should be able to get the tokenId/questionId and also the url associated with it that's why I made those changes - we need that connection

src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -9,63 +9,40 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+    mapping(uint256 => QuestionData) public questions;
+
+    constructor() {
         _tokenIdCounter.increment();
     }
 
-    // working standard metadata format:  Title, Description, Program
-    function safeMint(address to, string calldata uri) public onlyApi returns (uint256) {
+    function createQuestion(address author, string calldata uri) public onlyApi returns (uint256) {
         uint256 tokenId = _tokenIdCounter.current();
```
jdaly13, line 29: these are no longer tokens - think about renaming to fakeNFT or just questionCreatedInTime

src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -46,7 +46,7 @@ contract QuestionAPI is Ownable, NFTLocked {
         _costController.payForCreateQuestion(_msgSender());
 
         // Mint a new question
-        uint256 questionId = _question.safeMint(_msgSender(), uri);
+        uint256 questionId = _question.createQuestion(_msgSender(), uri);
```
jdaly13, line 49: there's an argument against naming every method the same as it becomes more difficult in debugging? something like mintQuestion IDK but we above you have tokenID but here it's question we shoudl remain consistent shoudl like hodl

src/contracts/Protocol/BountyQuestion.sol
gnomadic: yes!

src/contracts/Protocol/BountyQuestion.sol
gnomadic: I mean they still all have a unique ID?

src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -9,63 +9,40 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+        // QuestionData memory _question = QuestionData({tokenId: tokenId, url: uri});
+        authors[author].push(tokenId);
+        return tokenId;
     }
 
     function getAuthor(address user) public view returns (QuestionData[] memory) {
-        return authors[user];
+        uint256[] memory created = authors[user];
+
+        QuestionData[] memory ret = new QuestionData[](created.length);
```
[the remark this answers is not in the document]
gnomadic: it's a view function though!  I don't know how that works.  I'll do some reading.  I like the `QuestionData` struct you introduced, and I want that to be the primary data structure because this is a question.. ya know?  Like in a sql database we would have a table of these with a foreign key relationship to another table mapping authors.

src/contracts/Protocol/BountyQuestion.sol
gnomadic: I think this is covered by the view function below (and the discussion)

src/contracts/Protocol/QuestionAPI.sol
gnomadic: yeah you're right, we shoudl - fuck it we're still minting I'll rename

jdaly13 (approved):

PR #120, opened 2022-07-22
Update question states

thermohaline:
- Updated question states

### DIFF
--- a/src/contracts/Protocol/QuestionStateController.sol
+++ b/src/contracts/Protocol/QuestionStateController.sol
@@ -72,8 +72,8 @@ contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
         // TODO Unlock tokens for voting
     }
 
-    function setBadState(uint256 questionId) public onlyApi {
-        state[questionId] = STATE.BAD;
+    function setDisqualifiedState(uint256 questionId) public onlyApi {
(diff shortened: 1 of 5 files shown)


src/contracts/Protocol/QuestionStateController.sol
```diff
@@ -72,8 +72,8 @@ contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
         // TODO Unlock tokens for voting
     }
 
-    function setBadState(uint256 questionId) public onlyApi {
-        state[questionId] = STATE.BAD;
+    function setDisqualifiedState(uint256 questionId) public onlyApi {
```
jdaly13, line 75: This is fine but can't we just have a function that accepts any state and just set it that way - like setQuestionState(uint256 questionId, STATE questionState)
second argument would be a digit? but maybe it's not needed, you could set a question back to voting state???

thermohaline: I think that would be a good idea yes, however this PR serves only as an update for current logic with the STATE enum changed to what was discussed in the meeting. Think would be a good update to push in a different PR

jdaly13: makes sense - I'll create a low priority issue for it in the backlog

jdaly13 (approved):

PR #127, opened 2022-07-25
refactor to include votes

jdaly13:
refactor question model to include votes and refactor mapping for question by State

### DIFF
--- a/src/contracts/Protocol/QuestionAPI.sol
+++ b/src/contracts/Protocol/QuestionAPI.sol
@@ -121,10 +121,6 @@ contract QuestionAPI is Ownable, NFTLocked {
         _questionStateController.setDisqualifiedState(questionId);
     }
 
-    function setOffsetForQuestionState(uint256 newOffset) public onlyOwner {
-        _questionStateController.setOffset(newOffset);
-    }
-
     //------------------------------------------------------ Errors
     error ClaimsNotOpen();
     error NotAtBenchmark();
--- a/src/contracts/Protocol/QuestionAPI.sol
+++ b/src/contracts/Protocol/QuestionAPI.sol
@@ -42,11 +42,11 @@ contract QuestionAPI is Ownable, NFTLocked {
      * @return The question id
      */
     function createQuestion(string calldata uri, uint256 claimLimit) public returns (uint256) {
-        // Pay to create a question
-        _costController.payForCreateQuestion(_msgSender());
-
         // Mint a new question
         uint256 questionId = _question.mintQuestion(_msgSender(), uri);
+
+        // Pay to create a question
+        _costController.payForCreateQuestion(_msgSender(), questionId);
 
         // Initialize the question
         _questionStateController.initializeQuestion(questionId, uri);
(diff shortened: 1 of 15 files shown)


src/contracts/Protocol/QuestionStateController.sol
```diff
@@ -13,7 +13,7 @@ contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
     mapping(address => mapping(uint256 => bool)) public hasVoted;
     mapping(address => mapping(uint256 => uint256)) public questionIndex;
 
-    mapping(STATE => QuestionData[]) public questionByState;
+    mapping(STATE => mapping(uint256 => QuestionData)) public questionByState;
```
jdaly13, line 16: add additional mapping of questionID which we can get latest by calling getter on question API 
`await questionAPI.currentQuestionId();`

src/contracts/Protocol/QuestionStateController.sol
```diff
@@ -25,8 +25,8 @@ contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
      */
     function initializeQuestion(uint256 questionId, string calldata uri) public onlyApi {
         state[questionId] = STATE.VOTING;
-        QuestionData memory _question = QuestionData({questionId: questionId, url: uri, totalVotes: getTotalVotes(questionId)});
-        questionByState[STATE.VOTING].push(_question);
+        QuestionData memory _question = QuestionData({url: uri, totalVotes: 0});
```
jdaly13, line 28: when we initialize question we map it to the state of voting and the question ID so we can amend when we upvote

src/contracts/Protocol/QuestionStateController.sol
```diff
@@ -53,6 +53,8 @@ contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
 
         votes[questionId].totalVoteCount += amount;
 
+        QuestionData storage question = questionByState[STATE.VOTING][questionId];
```
jdaly13, line 56: now that we have the question data we can always amend votes to it

src/contracts/Protocol/QuestionStateController.sol
```diff
@@ -92,8 +94,14 @@ contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
         return votes[questionId].totalVoteCount;
     }
 
-    function getQuestionsByState(STATE currentState) public view returns (QuestionData[] memory) {
-        return questionByState[currentState];
+    function getQuestionsByState(STATE currentState, uint256 currentQuestionId) public view returns (QuestionData[] memory) {
```
jdaly13, line 97: we create a new array in memory because we know the amount we need to loop through by the second argument which is the latest or current question ID 
we loop backwards but have to create array from 0 so that's why we have i and j

jdaly13:
Before I get all questions I just need to get the latest question ID and then I can actually get all questions by state

src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -121,6 +121,10 @@ contract QuestionAPI is Ownable, NFTLocked {
         _questionStateController.setDisqualifiedState(questionId);
     }
 
+    function setOffsetForQuestionState(uint256 newOffset) public onlyOwner {
```
gnomadic, line 124: would this work if there are multiple people using the api at the same time?  should the offset be a method parameter instead?

gnomadic (approved):

PR #136, opened 2022-07-27
Beaut

thermohaline:
- Implementing the solidity style guide

### DIFF
--- a/src/contracts/Protocol/QuestionAPI.sol
+++ b/src/contracts/Protocol/QuestionAPI.sol
@@ -3,20 +3,49 @@ pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "./BountyQuestion.sol";
+
+// Interfaces
 import "./interfaces/IClaimController.sol";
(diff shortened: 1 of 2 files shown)


src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -3,20 +3,49 @@ pragma solidity 0.8.13;
...
+ * @notice This contract is an API for MetricsDAO that allows for interacting with questions & challenges.
+ */
+
 contract QuestionAPI is Ownable, NFTLocked {
     BountyQuestion private _question;
     IQuestionStateController private _questionStateController;
     IClaimController private _claimController;
     IActionCostController private _costController;
 
+    /// @notice Keeps track of the id of the most recent question created.
```
gnomadic, line 27: what's the difference between @notice and @dev?

thermohaline: From what I saw @notice is more general (solidity style docs: Explains to an end user what it does), where dev is more technical/extra detail

gnomadic (approved):

PR #141, opened 2022-07-28
Update claim logic and add tests

thermohaline:
- Fixed claim logic
- Added tests

### DIFF
--- a/src/contracts/Protocol/ClaimController.sol
+++ b/src/contracts/Protocol/ClaimController.sol
@@ -2,13 +2,53 @@
 pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts/access/Ownable.sol";
+
+// Interfaces
 import "./interfaces/IClaimController.sol";
+
+// Modifiers
 import "./modifiers/OnlyAPI.sol";
 
 contract ClaimController is Ownable, IClaimController, OnlyApi {
+    /// @notice Keeps track of claim limits per question
     mapping(uint256 => uint256) public claimLimits;
+
+    /// @notice maps answers to the question they belong to
     mapping(uint256 => mapping(address => Answer)) public answers;
+
+    /// @notice maps all claimers to a question
     mapping(uint256 => address[]) public claims;
+
+    //------------------------------------------------------ ERRORS
+
+    /// @notice Throw if user tries to claim a question that is past its limit
+    error ClaimLimitReached();
+
+    /// @notice Throw if a analyst tries to answer a question that it has not claimed
+    error NeedClaimToAnswer();
+
+    /// @notice Throw if analyst tries to claim a question multiple times
+    error AlreadyClaimed();
+
+    //------------------------------------------------------ STRUCTS
+
+    struct Answer {
+        STATE state;
+        address author;
+        string answerURL;
+        uint256 finalGrade;
+        string scoringMetaDataURI; // store heuristics and such on ipfs
+    }
+
+    // ------------------------------------------------------ ENUMS
+
+    enum STATE {
+        UNINT,
+        CLAIMED,
+        ANSWERED
+    }
+
+    // ------------------------------------------------------ FUNCTIONS
 
     /**
      * @notice Initializes a question to receive claims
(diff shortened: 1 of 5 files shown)


src/contracts/Protocol/ClaimController.sol
```diff
@@ -19,21 +59,24 @@ contract ClaimController is Ownable, IClaimController, OnlyApi {
...
-    function answer(uint256 questionId, string calldata answerURL) public onlyOwner {
-        if (answers[questionId][_msgSender()].state != STATE.CLAIMED) revert NeedClaimToAnswer();
-        answers[questionId][_msgSender()].answerURL = answerURL;
+    function answer(
+        address user,
+        uint256 questionId,
+        string calldata answerURL
+    ) public onlyOwner {
+        if (answers[questionId][user].state != STATE.CLAIMED) revert NeedClaimToAnswer();
+        answers[questionId][user].answerURL = answerURL;
```
jdaly13, line 77: Hmm are we not updating more properties in this function call or are we implementing separate function calls to update other states like finalGrade, scoringMetaDataURI?
or maybe answer isn't the best name for this IDK?

src/contracts/Protocol/ClaimController.sol
```diff
@@ -19,21 +59,24 @@ contract ClaimController is Ownable, IClaimController, OnlyApi {
...
+    function claim(address user, uint256 questionId) public onlyApi {
         if (claims[questionId].length >= claimLimits[questionId]) revert ClaimLimitReached();
+        if (answers[questionId][user].author == user) revert AlreadyClaimed();
 
-        claims[questionId].push(_msgSender());
-        Answer memory _answer = Answer({state: STATE.CLAIMED, author: _msgSender(), answerURL: "", scoringMetaDataURI: "", finalGrade: 0});
-        answers[questionId][_msgSender()] = _answer;
+        claims[questionId].push(user);
+        Answer memory _answer = Answer({state: STATE.CLAIMED, author: user, answerURL: "", scoringMetaDataURI: "", finalGrade: 0});
+        answers[questionId][user] = _answer;
```
jdaly13, line 68: I feel like we should add some JS Docs or something else to describe why we have these properties and why not sure where it's documented - looks fine but I don't remember discussing these properties

src/contracts/Protocol/ClaimController.sol
```diff
@@ -19,21 +59,24 @@ contract ClaimController is Ownable, IClaimController, OnlyApi {
...
-    function claim(uint256 questionId) public onlyOwner {
+    function claim(address user, uint256 questionId) public onlyApi {
         if (claims[questionId].length >= claimLimits[questionId]) revert ClaimLimitReached();
+        if (answers[questionId][user].author == user) revert AlreadyClaimed();
 
-        claims[questionId].push(_msgSender());
-        Answer memory _answer = Answer({state: STATE.CLAIMED, author: _msgSender(), answerURL: "", scoringMetaDataURI: "", finalGrade: 0});
-        answers[questionId][_msgSender()] = _answer;
+        claims[questionId].push(user);
+        Answer memory _answer = Answer({state: STATE.CLAIMED, author: user, answerURL: "", scoringMetaDataURI: "", finalGrade: 0});
```
jdaly13, line 67: How about an enum for the Grade so we know the available grades/score - I don't think any unsigned integer should be allowed for a grade

thermohaline: I like that idea, good call. Perhaps we could maybe cap it at 100 too. Will check what is most gas efficient in that too. However, I think that would be nice to iterate on and include in another PR. This PR mostly serves to make claims functional at the most basic level.

src/contracts/Protocol/ClaimController.sol
thermohaline: I agree, will look harder at this part when working on answering questions as well. For now again mostly just services to get the most basic version working.

src/contracts/Protocol/ClaimController.sol
thermohaline: Not too sure yet, currently this function is still untouched (also inaccessible because its still onlyOwner haha), once we get to work on figuring out the details of answering I think this will solve itself.

src/contracts/Protocol/ClaimController.sol
jdaly13: Sure can you just add a TODO and then we can merge in an iterate off it?

src/contracts/Protocol/ClaimController.sol
jdaly13: Same for below can you add a TODO to add JS Docs and some documentation for this so we can iterate off of

src/contracts/Protocol/ClaimController.sol
jdaly13: Like the other ones can you add a TODO for this as well

jdaly13:
I am approving but can you add some TODO's first and then merge

jdaly13 (approved):

PR #142, opened 2022-07-28
Fixes for voting

thermohaline:
- Fixed js tests
- Can no longer vote for own question
- Tweaked some stuff
- Added tests

### DIFF
--- a/src/contracts/Protocol/BountyQuestion.sol
+++ b/src/contracts/Protocol/BountyQuestion.sol
@@ -25,7 +25,7 @@ contract BountyQuestion is Ownable, OnlyApi {
         uint256 questionId = _questionIdCounter.current();
         _questionIdCounter.increment();
 
-        questions[questionId] = QuestionData({tokenId: questionId, url: uri});
+        questions[questionId] = QuestionData({author: author, tokenId: questionId, url: uri});
         authors[author].push(questionId);
         return questionId;
     }
@@ -40,7 +40,16 @@ contract BountyQuestion is Ownable, OnlyApi {
         return ret;
     }
 
+    function getAuthorOfQuestion(uint256 questionId) public view returns (address) {
+        return questions[questionId].author;
+    }
+
+    function getMostRecentQuestion() public view returns (uint256) {
+        return _questionIdCounter.current();
+    }
+
     struct QuestionData {
+        address author;
         uint256 tokenId;
         string url;
     }
--- a/src/contracts/Protocol/QuestionAPI.sol
+++ b/src/contracts/Protocol/QuestionAPI.sol
@@ -24,9 +24,6 @@ contract QuestionAPI is Ownable, NFTLocked {
     IClaimController private _claimController;
     IActionCostController private _costController;
 
-    /// @notice Keeps track of the id of the most recent question created.
-    uint256 public currentQuestionId;
-
     //------------------------------------------------------ ERRORS
 
     /// @notice Throw if analysts tries to claim a question that is not published.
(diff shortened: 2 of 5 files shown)


src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -108,6 +106,7 @@ contract QuestionAPI is Ownable, NFTLocked {
      * @param amount Metric amount to put behind the vote.
      */
     function upvoteQuestion(uint256 questionId, uint256 amount) public {
+        if (_question.getAuthorOfQuestion(questionId) == _msgSender()) revert CannotVoteForOwnQuestion();
```
jdaly13, line 109: Should it matter? - couldn't you potentially upvote your own question with many different addresses?  Just want to confirm this functionality

test/QuestionAPI-test.js
```diff
@@ -294,13 +290,13 @@ describe("Question API Contract", function () {
       const questionIDtx7 = await questionAPI.connect(xmetricaddr1).createQuestion("ipfs://last", 1);
       await questionIDtx7.wait();
 
-      const latestQuestion = await questionAPI.currentQuestionId();
+      const latestQuestion = await bountyQuestion.getMostRecentQuestion();
 
       const allquestionsByState = await questionStateController.getQuestionsByState(new BN(questionState.VOTING), latestQuestion, new BN(1000));
-      expect(allquestionsByState.length).to.equal(8);
+      expect(allquestionsByState.length - 1).to.equal(8);
```
jdaly13, line 296: If you removed a question shouldn't .length equal 7?
I tried to name tests so you see how many questions were created

test/QuestionAPI-test.js
```diff
@@ -294,13 +290,13 @@ describe("Question API Contract", function () {
...
 
       const allquestionsByStateAgain = await questionStateController.getQuestionsByState(new BN(questionState.VOTING), latestQuestion, new BN(4));
-      expect(allquestionsByStateAgain.length).to.equal(5);
+      expect(allquestionsByStateAgain.length - 1).to.equal(5);
```
jdaly13, line 299: same as above - if you remove question just change equal number - Better yet we should refactor to have that number be a variable like const allquestionsbyLength = 5; and then 
`expect(allquestionsByStateAgain.length).to.equal(5);`
Does that make sense?

src/contracts/Protocol/QuestionAPI.sol
thermohaline: Yeah exactly, I addressed that issue in the "MDAO through an exploit lense" thing I wrote. I think disallowing voting for your own question is beneficial, as we can assume a user who proposes a question likes the idea for their own question. It would then require every question proposer to always vote on their own question, requiring them to pay additional gas and lock additional METRIC. Given that anyone would have to do it, we might as well have nobody do it and start the like counter at 0. This way the METRIC that would be used to vote for your own question can be better spent voting for someone elses question.

test/QuestionAPI-test.js
thermohaline: Yeah you're right on both, that makes a lot more sense. My brain is tiny

src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -40,7 +40,16 @@ contract BountyQuestion is Ownable, OnlyApi {
         return ret;
     }
 
+    function getAuthorOfQuestion(uint256 questionId) public view returns (address) {
+        return questions[questionId].author;
```
gnomadic, line 44: if a question is created as a "challenge" by an admin the author is irrelevant - I don't think that matters and this code is good!  just a use-case to keep in mind

src/contracts/Protocol/QuestionAPI.sol
gnomadic: yeah - with the current design, it's OK if people buy metric in new wallets to upvote their questions -- if it's a good question, an admin will have to manually approve it to convert it into "PUBLISHED" - and if it's bad, then the metric will be slashed... and if metric is locked (and I hate to say, slashed) then it's good for the metric economy.

gnomadic (approved):

PR #145, opened 2022-08-01
Add and verify events

thermohaline:
- Events!

### DIFF
--- a/src/contracts/Protocol/QuestionAPI.sol
+++ b/src/contracts/Protocol/QuestionAPI.sol
@@ -45,7 +45,7 @@ contract QuestionAPI is Ownable, NFTLocked {
     event ChallengeCreated(uint256 indexed questionId, address indexed challengeCreator);
 
     /// @notice Emitted when a question is published.
-    event QuestionPublished(uint256 indexed questionId);
+    event QuestionPublished(uint256 indexed questionId, address indexed publisher);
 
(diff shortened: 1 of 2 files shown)


src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -130,6 +163,8 @@ contract QuestionAPI is Ownable, NFTLocked {
 
         // Publish the question
         _questionStateController.publish(questionId);
+
+        emit QuestionPublished(questionId);
```
gnomadic, line 167: should we throw the _msgSender() in here too?  for consistency and to track which admin hit publish?

src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -159,6 +198,8 @@ contract QuestionAPI is Ownable, NFTLocked {
      */
     function disqualifyQuestion(uint256 questionId) public onlyOwner {
         _questionStateController.setDisqualifiedState(questionId);
+
+        emit QuestionDisqualified(questionId);
```
gnomadic, line 202: same question as above

gnomadic (approved):

src/contracts/Protocol/QuestionAPI.sol
thermohaline: agreed

PR #146, opened 2022-08-02
Update vault logic v1

thermohaline:
First vault revamp update.

The problem we were facing is the fact that we have multiple stages where costs are processed. It can be mapped like this:

### Creation and voting

- Lock metric for creation (per qId)
- Lock metric for voting (per qId)

Unlock on publish

### Claiming & answering

- Lock metric for claiming (per qId)
- Lock metric for answering (per qId)

Unlock on end of submission period

### Reviewing

- Lock metric to review (per qId)

Unlock at end of review period

In the current workings this would mean a separate deposit & withdrawal function for each action.

This vault update does the following:
- Single deposit & withdraw function
- Deposits are managed per stage
- Withdrawals are managed per stage

Where we pass stages [0, 1, 2] for Creation and voting, Claiming and answering and finally Reviewing.

Each stage is independently accounted for per question.

### DIFF
--- a/src/contracts/Protocol/ActionCostController.sol
+++ b/src/contracts/Protocol/ActionCostController.sol
@@ -2,28 +2,28 @@
 pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts/access/Ownable.sol";
-import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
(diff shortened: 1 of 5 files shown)


jdaly13 (approved):
Looks mostly good to me - for tests either update test or remove or put a TODO to update

thermohaline:
Yep I opened an issue on it

PR #151, opened 2022-08-03
143 vote for your own question first vote is free

thermohaline:
- Cleaned up voting
- Added voting cost
- Fixed tests
- Verified vault logic
- Fixed slight vault issue
- Bring voting data to a central place
- Made first vote free

### DIFF
--- a/src/contracts/Protocol/ActionCostController.sol
+++ b/src/contracts/Protocol/ActionCostController.sol
@@ -32,23 +32,18 @@ contract ActionCostController is Ownable, OnlyApi, IActionCostController {
             the contract.
     * @param _user The address of the user who wants to pay for creating a question.
     */
-    function payForCreateQuestion(address _user, uint256 _questionId) external onlyApi {
-        // Checks
-        // Effects
-        // Interactions
-        vault.lockMetric(_user, createCost, _questionId, 0);
+    function payForCreateQuestion(address _user, uint256 questionId) external onlyApi {
+        vault.lockMetric(_user, createCost, questionId, 0);
     }
 
     /**
     * @notice Makes a user pay for voting on a question. 
             We transfer the funds from the user executing the function to 
             the contract.
-    * @param _user The address of the user who wants to pay for voting on a question.
+    * @param user The address of the user who wants to pay for voting on a question.
     */
-    function payForVoting(address _user) external onlyApi {
-        // Checks
-        // Effects
-        // Interactions
+    function payForVoting(address user, uint256 questionId) external onlyApi {
+        vault.lockMetric(user, voteCost, questionId, 0);
     }
 
     // ------------------------------- Getter
--- a/src/contracts/Protocol/QuestionStateController.sol
+++ b/src/contracts/Protocol/QuestionStateController.sol
@@ -2,18 +2,19 @@
 pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts/access/Ownable.sol";
+
+// Interfaces
 import "./interfaces/IQuestionStateController.sol";
+
+// Modifiers
 import "./modifiers/OnlyAPI.sol";
 
 contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
-    mapping(uint256 => QuestionVote) public votes;
-    mapping(uint256 => STATE) public state;
-
     // Mapping for all questions that are upvoted by the user?
     mapping(address => mapping(uint256 => bool)) public hasVoted;
     mapping(address => mapping(uint256 => uint256)) public questionIndex;
 
-    mapping(uint256 => QuestionData) public questionByState;
+    mapping(uint256 => QuestionStats) public questionByState;
 
     //TODO mapping     mapping(STATE => uint256[]) public questionState;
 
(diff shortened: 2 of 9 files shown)


src/contracts/Protocol/ActionCostController.sol
```diff
@@ -32,23 +32,18 @@ contract ActionCostController is Ownable, OnlyApi, IActionCostController {
...
             We transfer the funds from the user executing the function to 
             the contract.
-    * @param _user The address of the user who wants to pay for voting on a question.
+    * @param user The address of the user who wants to pay for voting on a question.
     */
-    function payForVoting(address _user) external onlyApi {
-        // Checks
-        // Effects
-        // Interactions
+    function payForVoting(address user, uint256 questionId) external onlyApi {
```
jdaly13, line 45: like creating a question we'll need to approve this transaction before as well

src/contracts/Protocol/ActionCostController.sol
```diff
@@ -32,23 +32,18 @@ contract ActionCostController is Ownable, OnlyApi, IActionCostController {
...
+        vault.lockMetric(user, voteCost, questionId, 0);
```
jdaly13, line 46: Is there a reason to have two functions that take in same params - I think it's fine because we can change each API as needed even though now they are identical

src/contracts/Protocol/QuestionStateController.sol
```diff
@@ -24,35 +25,34 @@ contract QuestionStateController is IQuestionStateController, Ownable, OnlyApi {
...
-        votes[questionId].votes.push(_vote);
+        QuestionStats storage _question = questionByState[questionId];
+        _question.totalVotes += 1;
 
         hasVoted[_user][questionId] = true;
-        questionIndex[_user][questionId] = votes[questionId].votes.length - 1;
-
-        votes[questionId].totalVoteCount += amount; // TODO Lock tokens for voting include safeTransferFrom
+        _question.voters.push(_user);
+        questionIndex[_user][questionId] = _question.voters.length - 1;
```
jdaly13, line 54: I guess I'm sort of confused on the purpose of this or what this does

src/contracts/Protocol/Vault.sol
```diff
@@ -204,6 +206,26 @@ contract Vault is Ownable, OnlyCostController {
         return lockedMetric[questionId][stage][user];
     }
 
+    function getLockedMetricByQuestion(uint256 questionId) public view returns (uint256) {
+        return lockedMetricByQuestion[questionId];
+    }
+
+    function getUserFromProperties(
+        uint256 questionId,
+        uint256 stage,
```
jdaly13, line 215: what is stage?

test/QuestionAPI-test.js
```diff
@@ -106,6 +106,7 @@ describe("Question API Contract", function () {
     it("the factory should create questions", async function () {
       // create question
       await xmetric.connect(xmetricaddr1).approve(vault.address, ethers.utils.parseEther("30"));
+      await xmetric.connect(xmetricaddr2).approve(vault.address, ethers.utils.parseEther("30"));
```
jdaly13, line 109: not a biggie I would jut change number to differentiate - also add a TODO to approve amount below so we can test rejection of upvote

jdaly13: also line 88 and 108 are same - I think we need to fix that, that might of been my bad when I was cleaning up

test/QuestionAPI-test.js
```diff
@@ -167,10 +168,11 @@ describe("Question API Contract", function () {
 
     it("the facotry should enable voting once a question is created", async () => {
```
jdaly13, line 169: good stuff

jdaly13:
Great work on this - I made a few comments but I think it's good to go - I think we should update JS tests but we can set a TODO for that

jdaly13 (approved):

src/contracts/Protocol/Vault.sol
gnomadic: as a follow up to this one - should we use enums instead of uints?  What's nice about enums is it's more self-explanatory - but what's nice about uints is that we can just make em up as we go.. as we talked about in our product discussion we might wanna mess around with stages for xMetric...  what do ya'll think?

src/contracts/Protocol/ActionCostController.sol
thermohaline: Good shout, let me take a look at simplifying this

src/contracts/Protocol/QuestionStateController.sol
thermohaline: yep!

src/contracts/Protocol/Vault.sol
thermohaline: Enums it is! I like that, we can always just add another stage in the enum

test/QuestionAPI-test.js
thermohaline: Will do

src/contracts/Protocol/ActionCostController.sol
thermohaline: Made a TODO for it

PR #164, opened 2022-08-09
Convert to enums

thermohaline:


src/contracts/Protocol/Vault.sol
```diff
@@ -156,7 +159,9 @@ contract Vault is Ownable, OnlyCostController {
             metric.transfer(_msgSender(), toWithdraw);
 
             emit Withdraw(_msgSender(), toWithdraw);
-        } else if (stage == 1) {
+        } else if (stage == STAGE.UNVOTE) {
```
gnomadic, line 162: not yet - but as this if/else tree grows we might want to refactor the functionality into their own methods

src/contracts/Protocol/Vault.sol
```diff
@@ -131,15 +134,15 @@ contract Vault is Ownable, OnlyCostController {
      * @param questionId The question id
      * @param stage The stage for which the user is withdrawing metric from a question.
      */
-    function withdrawMetric(uint256 questionId, uint256 stage) external {
+    function withdrawMetric(uint256 questionId, STAGE stage) external {
         // Checks if Metric is withdrawn for a valid stage.
-        if (stage >= 3) revert InvalidStage();
+        if (uint8(stage) >= 4) revert InvalidStage();
```
gnomadic, line 139: is there a way to avoid hardcoding 4? I don't know of any

gnomadic (approved):

src/contracts/Protocol/Vault.sol
thermohaline: Yep!

PR #166, opened 2022-08-09
showcasing data refactoring direction

gnomadic:


src/contracts/Protocol/Structs/QuestionData.sol
```diff
@@ -0,0 +1,13 @@
+//SPDX-License-Identifier: MIT
+pragma solidity 0.8.13;
+
+import "../interfaces/IQuestionStateController.sol";
+
+struct QuestionData {
```
jdaly13, line 6: so this questionData would be used by question state controller as well?

src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -48,7 +57,11 @@ contract BountyQuestion is Ownable, OnlyApi {
         return _questionIdCounter.current();
     }
 
-    struct QuestionData {
+    function getQuestionData(uint256 questionId) public view returns (QuestionData memory) {
+        return questionData[questionId];
+    }
+
+    struct QuestionMetaData {
```
jdaly13, line 64: to keep consistent probably change url to uri as I had to change this on front end as well

jdaly13 (commented):
This looks good but are we integrating these changes into the question State controller as well in this PR?

thermohaline (approved):

PR #167, opened 2022-08-10
PD-8 users are reimbursed on unvoting

thermohaline:
- Used the wrong branch on accident but results are the same

### DIFF
--- a/src/contracts/Protocol/Vault.sol
+++ b/src/contracts/Protocol/Vault.sol
@@ -43,8 +43,12 @@ contract Vault is Ownable, OnlyCostController {
     error NoMetricDeposited();
     /// @notice Throw if user tries to lock Metric for a question that has a different state than UNINT.
     error QuestionHasInvalidStatus();
+    /// @notice Throw if user tries to claim Metric for unvoting on a question that is not in the VOTING state.
+    error QuestionNotInVoting();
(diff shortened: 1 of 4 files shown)


src/contracts/Protocol/Vault.sol
```diff
@@ -137,37 +141,45 @@ contract Vault is Ownable, OnlyCostController {
...
         } else {
             // if (reviewPeriod == active) revert ReviewPeriodActive();
         }
     }
 
+    function vaultAccounting(uint256 questionId, STAGE stage) internal {
+        if (_msgSender() != lockedMetric[questionId][stage][_msgSender()].user) revert NotTheDepositor();
+        if (lockedMetric[questionId][stage][_msgSender()].status != STATUS.DEPOSITED) revert NoMetricDeposited();
+
+        uint256 toWithdraw = lockedMetric[questionId][stage][_msgSender()].amount;
```
gnomadic, line 169: could this get funky if the user has lockedMetric for different reasons?  like when they create a question it gets an auto-upvote -if they unvote, would it reimburse their creation fees?

thermohaline: Yep you're right, good catch. Fixing.

thermohaline:
Fixed

gnomadic (approved):

PR #168, opened 2022-08-10
Pd 18 use library pattern to hold enums and structs

thermohaline:


jdaly13:
@gnomadic and @thermohaline  - we have to be mindful when changing enums that we need to change the enums in the front end as well because there are no getters for enums in solidity and so they need to be mapped in the front end the same way 
for example 
`export const questionStateEnum = {
    UNINIT: 0,
    VOTING: 1,
    PUBLISHED: 2,
    IN_GRADING: 3,
    COMPLETED: 4,
    CANCELLED: 5,
    BAD: 6,
};`

gnomadic:
https://github.com/MetricsDAO/xyz/pull/37 will take care of the enums on the client side!

gnomadic (approved):

PR #169, opened 2022-08-11
Pd 27 question claiming

thermohaline:


src/contracts/Protocol/ClaimController.sol
```diff
@@ -60,20 +54,29 @@ contract ClaimController is Ownable, IClaimController, OnlyApi {
...
         answers[questionId][user] = _answer;
     }
 
+    function releaseClaim(address user, uint256 questionId) public onlyApi {
+        if (answers[questionId][user].author != user) revert NoClaimToRelease();
+
+        answers[questionId][user].state = CLAIM_STATE.RELEASED;
+        answers[questionId][user].author = address(0);
+
+        --claimCounts[questionId];
```
gnomadic, line 71: is this `--` syntax the same `delete`?

gnomadic: HA I see it now, the mapping stores an uint.

src/contracts/Protocol/ClaimController.sol
```diff
@@ -6,13 +6,22 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+
 // Modifiers
 import "./modifiers/OnlyAPI.sol";
 
 contract ClaimController is Ownable, IClaimController, OnlyApi {
     /// @notice Keeps track of claim limits per question
     mapping(uint256 => uint256) public claimLimits;
 
+    /// @notice Keeps track of claim counts per question
+    mapping(uint256 => uint256) public claimCounts;
```
gnomadic, line 23: wouldn't

claimCounts[X] == answers[x].length

?  Do we need a second mapping for this?

gnomadic (approved):

PR #172, opened 2022-08-22
Test revamp

thermohaline:
invariant testing setup


foundry.toml
```diff
@@ -12,4 +12,20 @@ remappings = [
...
+invariant_call_override = false  # Override calls
+invariant_fail_on_revert = false  # Fail the test if the contract reverts
+
+[profile.local]
+fuzz_runs = 100
+
+[profile.deep]
+fuzz_runs = 500
+
+[profile.super_deep]
```
gnomadic, line 30: lol

src/test/Claiming/Claiming.t.sol
```diff
@@ -0,0 +1,257 @@
...
+    BountyQuestion _bountyQuestion;
+    ClaimController _claimController;
+    ActionCostController _costController;
+    QuestionStateController _questionStateController;
+    Vault _vault;
+    NFT _mockAuthNFTManager;
+    NFT _mockAuthNFTAdmin;
+
+    /// @notice Emitted when a question is created.
+    event QuestionCreated(uint256 indexed questionId, address indexed creator);
```
gnomadic, line 44: can these be imported instead of recreated?

gnomadic (approved):

src/test/Claiming/Claiming.t.sol
thermohaline: Yep, will do that together with the rest!

PR #173, opened 2022-08-29
Foundry deploy

thermohaline:


src/contracts/Protocol/Extra/MockAuthNFT.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.13;
 import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
-contract NFT is ERC721, Ownable {
+contract MockAuthNFT is ERC721, Ownable {
```
jdaly13, line 7: This could be fine - I just don't like the idea of naming it Mock - which is a term mostly used in development for Testing - maybe renaming it to quasi or even pseudo or imitation

jdaly13: Also confused why we need  a new contract for deployment purposes?

PR #174, opened 2022-08-31
Challenge proposal, metric burning

thermohaline:


src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -109,11 +112,31 @@ contract QuestionAPI is Ownable, NFTLocked, FunctionLocked {
         return questionId;
     }
 
+    /**
+     * @notice Creates a challenge.
+     * @param uri The IPFS hash of the challenge.
+     * @return The challenge id
+     */
+    function proposeChallenge(string calldata uri) public returns (uint256) {
```
jdaly13, line 120: confused by this - createChallenge seems to be gated but here we are doing something similar skipping the question creation going straight to minting without it being gated?

I'm also confused on the functionality where a question becomes a challenge 
What's the point of minting a question in createQuestion if both methods of creating and proposing a challenge to do the same thing - we are missing the step 
`function instantiateChallengeFromQuestion`

src/contracts/Protocol/Vault.sol
```diff
@@ -124,6 +124,10 @@ contract Vault is Ownable, OnlyCostController {
         depositAccounting(user, amount, questionId, stage);
     }
 
+    function burnMetric(address user, uint256 amount) external onlyCostController {
+        metric.transferFrom(user, treasury, amount);
```
jdaly13, line 128: so the treasury is just some address we what is the address are we truly burning these tokens?

src/contracts/Protocol/QuestionAPI.sol
jdaly13: so when you burn you just send 1000 metric to the treasury, the treasury being a burnable address

jdaly13:
> Explanation:
> I had jumped in a call with Meg to verift some of the requirements for MVP2. Because they want to utilize xMETRIC a bit more they came up with the idea of being able to burn 1000 xMETRIC in order to skip the voting stage of a question. We therefore do not “create a question”, but rather “propose a challenge”. This challenge (just like a sufficiently upvoted question) has to then be accepted by a holder of the NFT admin role, who then decides whether or not to publish the question. To allow this I added actionCost[ACTION.CHALLENGE_BURN] so we can change the amount of xMETRIC to burn to propose a challenge at any point.
> To clarify on your comments (I don’t blame you, I was super confused by this as well):
> - We used to have a gated createChallenge function that could be used by program managers to directly create a challenge. We keep this functionality, but in addition to that add a “proposeChallenge” function. This can be called by anyone in order to “skip the line” on voting. However, the createChallenge function automatically publishes the challenge, whereas the proposeChallenge function still requires approval of an admin before being published.
> - The reason why we need to use burnForAction instead of payForAction is because we are wanting to burn the tokens. The payForAction function automatically vaults the money, which is not what we want in this case -- the xMETRIC should not be recoverable. I have for now, as you noticed, set it to a treasury address. We are supposed to burn the tokens but the ERC20 standard doesn’t allow transactions to the 0 address. We could either implement ERC20burnable, or just send it to an alternative burn address such as 0x000000...0dEaD. But yes, as you correctly said, when you call the burnForAction now, it sends 1000 xMETRIC to the treasury address, where onlyOwner is able to set the treasury address.
> - I agree with your call on renaming “publishChallengeFromQuestion”, gives a lot more clarity in the contract.

Thanks for clarifying @thermohaline - can you please rename publishChallenge function and also put a todo for burning Metric to either a burnable address manually or inherit from a burnable contract

PR #175, opened 2022-08-31
Total vote fix

thermohaline:


jdaly13 (commented):
Looks good but we should remove 
```uint256 totalVotes;``` from the Votes struct as it looks like it's not being utilized anymore the source of truth has moved to bountyQuestion it is presumed

jdaly13:
let't remove totalVotes from struct and merge in - I will approve for now just make sure to update before merging

jdaly13 (approved):
please see other comment

PR #176, opened 2022-09-01
PD-41 added token barring

thermohaline:


src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -115,13 +115,17 @@ contract QuestionAPI is Ownable, NFTLocked, FunctionLocked {
      * @param claimLimit The limit for the amount of people that can claim the challenge
      * @return questionId The question id
      */
-    function createChallenge(string calldata uri, uint256 claimLimit) public onlyHolder(PROGRAM_MANAGER_ROLE) returns (uint256) {
+    function createChallenge(
+        string calldata uri,
+        uint256 claimLimit,
```
jdaly13, line 120: can we make these numbers smaller?  Also should we set a max claimLimit or max threshold?
also can you add param documentation for threshold above function

src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -163,10 +167,14 @@ contract QuestionAPI is Ownable, NFTLocked, FunctionLocked {
      * @param claimLimit The amount of claims per question.
      */
 
-    function publishQuestion(uint256 questionId, uint256 claimLimit) public onlyHolder(ADMIN_ROLE) {
+    function publishQuestion(
+        uint256 questionId,
```
jdaly13, line 171: same comment as above

src/contracts/Protocol/modifiers/TokenBar.sol
```diff
@@ -0,0 +1,28 @@
...
+import "@openzeppelin/contracts/interfaces/IERC20.sol";
+import "../interfaces/IClaimController.sol";
+
+abstract contract TokenBar is Ownable {
+    address public metricToken;
+
+    /// @notice Throw if user tries to claim with balance lower than threshold
+    error NotEnoughTokens();
+
+    function setMetricToken(address _metricToken) public onlyOwner {
```
jdaly13, line 16: WE have this same function in 3 different places now - the Vault and costController and now here - can we combine these somehow?

src/contracts/Protocol/ClaimController.sol
```diff
@@ -49,11 +53,16 @@ contract ClaimController is Ownable, IClaimController, OnlyApi {
...
      * @param claimLimit The limit for the amount of people that can claim the question
      */
-    function initializeQuestion(uint256 questionId, uint256 claimLimit) public onlyApi {
+    function initializeQuestion(
+        uint256 questionId,
+        uint256 claimLimit,
+        uint256 threshold
+    ) public onlyApi {
         claimLimits[questionId] = claimLimit;
+        metricThresholds[questionId] = threshold;
```
jdaly13, line 62: need to add documentation for this param

thermohaline: Will add documentation on approval, just in case anything changes etc

src/contracts/Protocol/QuestionAPI.sol
thermohaline: Is there any reason to make the numbers smaller? And we do set both claimLimit and threshold in `        _claimController.initializeQuestion(questionId, claimLimit, threshold);
`

src/contracts/Protocol/modifiers/TokenBar.sol
thermohaline: Yeah will look into that, single source of truth for metric address seems like a better approach

src/contracts/Protocol/QuestionAPI.sol
jdaly13: meaning do they need to be unit256? not a biggie just a question

src/contracts/Protocol/ClaimController.sol
jdaly13: I meant param documentation in the code itself for example
```* @param claimLimit The amount of claims per question.```

File: src/contracts/Protocol/QuestionAPI.sol (hunk 2 above)
Comment (jdaly13) [reply src/contracts/Protocol/QuestionAPI.sol:171]: ```@param threshhold```

**Review by jdaly13 — CHANGES_REQUESTED** [review]

File: src/contracts/Protocol/QuestionAPI.sol (hunk 1 above)
Comment (thermohaline) [reply src/contracts/Protocol/QuestionAPI.sol:120]: Ah yes fair, my understanding was that uint256 is more gas friendly. Uint8 does uint8(uint256(1)), therefore costing more gas.

File: src/contracts/Protocol/ClaimController.sol (hunk 4 above)
Comment (thermohaline) [reply src/contracts/Protocol/ClaimController.sol:62]: Ohhh yes my bad -- adding right away

**Review by gnomadic — APPROVED** [review]

File: src/contracts/Protocol/ClaimController.sol [hunk 5]
```diff
@@ -14,8 +14,9 @@ import "./Structs/AnswerStruct.sol";
 
 // Modifiers
 import "./modifiers/OnlyAPI.sol";
+import "./modifiers/TokenBar.sol";
 
-contract ClaimController is Ownable, IClaimController, OnlyApi {
```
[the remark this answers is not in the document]
thermohaline: We dont seem to be using it anywhere, but will fix if it breaks anything post merge

jdaly13 (approved):

PR #178, opened 2022-09-05
Correct burning of tokens

thermohaline:


jdaly13:
Just merge in 179 which has these changes and then some

PR #179, opened 2022-09-05
Pd 46 check that nothing weird happens with nonexistant q ids

thermohaline:


jdaly13 (approved):
great Work on this!!!!!!!!

PR #180, opened 2022-09-09
Pd 43 single source truth controllers

thermohaline:


src/contracts/Protocol/BountyQuestion.sol
```diff
@@ -3,13 +3,19 @@ pragma solidity 0.8.13;
...
+// Modifiers
 import "./modifiers/OnlyAPI.sol";
 import "./modifiers/OnlyStateController.sol";
+
+// Structs
 import "./Structs/QuestionData.sol";
-import "./interfaces/IBountyQuestion.sol";
 
 /// @custom:security-contact contracts@metricsdao.xyz
-contract BountyQuestion is IBountyQuestion, Ownable, OnlyApi, OnlyStateController {
```
gnomadic, line 12: so you don't actually have to to implement the interface -- but other contracts will use the interface to call into this one (so that we don't have to include the entire file or write the ABI by hand)

If we don't do implement these interfaces, we need to be confident that a change in any "public" method signatures will be updated in other files (because the interface for the consumer could be out of sync with the implementation)

thermohaline: That should be fine though right, we could just write some test coverage so that they'd fail if they go out of sync

gnomadic (approved):

PR #183, opened 2022-09-12
create test public question rename mockAuthNFT to PseudoMockNFT

jdaly13:


jdaly13:
Hardhat requires in JS tests to have the contract name be the same name as file

thermohaline (approved):

PR #184, opened 2022-09-13
Vault rework to be controlled from QuestionAPI

thermohaline:


src/contracts/Protocol/QuestionAPI.sol
```diff
@@ -265,6 +273,12 @@ contract QuestionAPI is Ownable, NFTLocked, FunctionLocked {
         emit QuestionDisqualified(questionId, _msgSender());
     }
 
+    function withdrawFromVault(uint256 questionId, STAGE stage) public {
```
jdaly13, line 276: Why not just have the two above functions call this function and send in the appropriate stage - or I guess what's the use case for this one?

src/contracts/Protocol/Vault.sol
```diff
@@ -122,10 +123,15 @@ contract Vault is Ownable, OnlyCostController {
...
      * @param stage The stage for which the user is withdrawing metric from a question.
      */
-    function withdrawMetric(uint256 questionId, STAGE stage) external {
+    function withdrawMetric(
+        address user,
+        uint256 questionId,
+        STAGE stage
+    ) external onlyApi {
         // Checks if Metric is withdrawn for a valid stage.
         if (uint8(stage) >= 5) revert InvalidStage();
```
jdaly13, line 136: I know the change wasn't made in this one but 5 feels arbitrary why not > STAGE.REVIEW

thermohaline: Good idea

jdaly13 (commented):
This looks good my question is - when you withdraw metric for example you can unvote or release a claim but what about the other stages? like
    CREATE_AND_VOTE,
    CLAIM_AND_ANSWER 
    REVIEW
   
Or will these be implemented later?

thermohaline:
> This looks good my question is - when you withdraw metric for example you can unvote or release a claim but what about the other stages? like CREATE_AND_VOTE, CLAIM_AND_ANSWER REVIEW
> 
> Or will these be implemented later?

The other stages as of now you manually enter upon withdrawing. This is because the actions voting, as well as creating questions both fall within the CREATE_AND_VOTE stage. Because we "dynamically" want to unlock tokens through a question life span, we unlock per stage like this. It is not possible to do this automatically as we'd have to loop through all users that have interacted with the question and update a withdraw-able allowance or something

jdaly13:
> > This looks good my question is - when you withdraw metric for example you can unvote or release a claim but what about the other stages? like CREATE_AND_VOTE, CLAIM_AND_ANSWER REVIEW
> > Or will these be implemented later?
> 
> The other stages as of now you manually enter upon withdrawing. This is because the actions voting, as well as creating questions both fall within the CREATE_AND_VOTE stage. Because we "dynamically" want to unlock tokens through a question life span, we unlock per stage like this. It is not possible to do this automatically as we'd have to loop through all users that have interacted with the question and update a withdraw-able allowance or something

Makes sense thanks for clarifying - so essentially when you release a claim or unvote those are the only two cases in which you would automatically withdraw metric

PR #187, opened 2022-09-14
create tests and costs for publish and create challenge

jdaly13:
Default to 0

### DIFF
--- a/src/contracts/Protocol/ActionCostController.sol
+++ b/src/contracts/Protocol/ActionCostController.sol
@@ -29,10 +29,14 @@ contract ActionCostController is Ownable, OnlyApi, IActionCostController {
         actionCost[ACTION.VOTE] = 1e18;
         actionCost[ACTION.CLAIM] = 1e18;
         actionCost[ACTION.CHALLENGE_BURN] = 1000e18;
+        actionCost[ACTION.CHALLENGE_CREATE];
+        actionCost[ACTION.PUBLISH];
 
         actionStage[ACTION.CREATE] = STAGE.CREATE_AND_VOTE;
         actionStage[ACTION.VOTE] = STAGE.CREATE_AND_VOTE;
         actionStage[ACTION.CLAIM] = STAGE.CLAIM_AND_ANSWER;
+        actionStage[ACTION.CHALLENGE_CREATE] = STAGE.REVIEW;
+        actionStage[ACTION.PUBLISH] = STAGE.REVIEW;
     }
 
     // ------------------------------------------------------ FUNCTIONS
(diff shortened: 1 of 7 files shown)


src/contracts/Protocol/ActionCostController.sol
```diff
@@ -29,10 +29,14 @@ contract ActionCostController is Ownable, OnlyApi, IActionCostController {
         actionCost[ACTION.VOTE] = 1e18;
         actionCost[ACTION.CLAIM] = 1e18;
         actionCost[ACTION.CHALLENGE_BURN] = 1000e18;
+        actionCost[ACTION.CHALLENGE_CREATE];
+        actionCost[ACTION.PUBLISH];
 
         actionStage[ACTION.CREATE] = STAGE.CREATE_AND_VOTE;
         actionStage[ACTION.VOTE] = STAGE.CREATE_AND_VOTE;
         actionStage[ACTION.CLAIM] = STAGE.CLAIM_AND_ANSWER;
+        actionStage[ACTION.CHALLENGE_CREATE] = STAGE.REVIEW;
```
jdaly13, line 38: I wasn't sure if this was correct - If I don't state an actionStage foundry tests fail but not JS interestingly enough

thermohaline: Im assuming this is because the JS does not have vault test coverage. This would store the payments for question creation in the vault under REVIEW, meaning funds would currently be lost. We can probably put both under CREATE_AND_VOTE.

jdaly13: vault coverage is next on my list :-)

jdaly13: updated

src/contracts/Protocol/Vault.sol
```diff
@@ -76,7 +76,8 @@ contract Vault is Ownable, OnlyCostController, OnlyApi, VaultEventsAndErrors {
         // Checks if METRIC is locked for a valid stage.
         if (uint8(stage) >= 5) revert InvalidStage();
         // Checks if there has not been a deposit yet
-        if (lockedMetric[questionId][stage][user].status != STATUS.UNINT) revert QuestionHasInvalidStatus();
+        // TODO remove amount!=0 possibly
+        if (lockedMetric[questionId][stage][user].status != STATUS.UNINT && amount != 0) revert QuestionHasInvalidStatus();
```
jdaly13, line 80: I think this works for now - but I'm thinking we should change tests as well because a user shouldn't be creating a question and then publishing it

thermohaline: Wouldn't introducing amount != 0 mean that we cannot set cost to 0? Also think published should be a separate state, otherwise they conflict (explained in slack)

jdaly13: no it means you can't have a question that costs money to create and cost money to publish

jdaly13: that was the intention but I updated this to not to do that

thermohaline (approved):

